PR #2, opened 2017-09-19
Fix: Removed all hardcoded lables and text message

nilesh-more:


nilesh-more:
@harishgilimi ref this https://github.com/project-sunbird/sunbird-portal/pull/7

PR #109, opened 2017-11-27
Sb 945 model service

manjudr:


src/app/helpers/announcement/ObjectStore/ObjectStoreRest.js
```diff
@@ -80,63 +87,41 @@ class ObjectStoreRest extends ObjectStore {
...
+                    body: {
+                        request: {
+                            'tableName': data.table,
+                            'documentName': data.table,
+                            'payload': data.values
+                        }
+                    },
+                    json: true
+                }
                 let result = await (this.httpService(options))
```
kochhar: why does this use its own httpservice? why not use the http wrapper which is used in the notification service also?

src/app/helpers/announcement/controller.js
```diff
@@ -17,14 +16,11 @@ let ApiInterceptor = require('sb_api_interceptor')
...
 
-  constructor() {
-    //table name should be same as the name in database table
-    let tableMapping = {
-      'announcement': AnnouncementModel,
-      'announcementtype': AnnouncementTypeModel,
-      'attachment': AttachmentModel,
-      'metrics': MetricsModel
-    }
+  constructor({metrics, announcement, announcementtype } = {}) {
```
kochhar: this constructor needs documentation to explain the object passed in and the properties which they must have.

src/app/helpers/announcement/model/AnnouncementCreateModel.js
```diff
@@ -0,0 +1,83 @@
...
+ */
+
+let BaseModel = require('./BaseModel.js')
+let Joi = require('joi')
+
+/**
+ * Announcement create model validation
+ * @extends BaseModel
+ */
+class AnnouncementModel extends BaseModel {
```
kochhar: this should have two additional methods: fromApi and fromDb

src/app/helpers/announcement/ObjectStore/ObjectStoreRest.js
manjudr: This is not in the part of notification service we have done this clean up in announcement model service task

kochhar: great.

PR #186, opened 2018-01-03
Issue #SB-1165 feat: query service implementation

nilesh-more:


harishgilimi:
@nilesh-more query service returns instance of data source it won't call the methods directly

PR #241, opened 2018-02-21
Issue #SB-1376 fix:added validation in profile page

vbrajath:
@harishgilimi


src/app/private/scripts/controllers/profileController.js
```diff
@@ -721,6 +790,7 @@ angular.module('playerApp')
       profile.getSkills()
 
       profile.openAddSkillModal = function () {
+        $('#addSkill').attr('maxlength', 10)
```
harishgilimi: Move this to config "10"

harishgilimi:
Do complete testing we should not get any issues

vbrajath:
Okay harish, I will do.

PR #299, opened 2018-03-21
SB-1171 : Announcement : When user selects multiple files with the combination of valid and invalid file format, valid one also does not gets uploaded

souravdey091:


src/app/private/scripts/factories/fileUploadFactory.js
```diff
@@ -86,8 +87,7 @@ angular.module('playerApp')
               template: 'qq-template-manual-trigger',
               autoUpload: true,
               paramsInBody: true,
-              debug: false,
-              stopOnFirstInvalidFile: false,
+              debug: false,
```
gsbajaj72: Please remove the trailing spaces and update PR

PR #396, opened 2018-04-17
Issue #SB-2157: fix reverted config, getter fn for dims, channel

revathitarento:


src/app/migration/src/src/app/modules/workspace/components/create-editors/collection-editor/collection-editor.component.ts
```diff
@@ -344,78 +345,7 @@ export class CollectionEditorComponent implements OnInit, AfterViewInit {
     const editorConfig = [];
     switch (type) {
       case 'Course':
-        editorConfig.push({
```
anoophm: why was this removed

PR #462, opened 2018-05-03
Issue #SB-3018 feat: Limited published component and sharelink component

amantarento:
Issue #SB-3018 feat: Limited published component and sharelink component


harishgilimi:
@amantarento  cross check all the files once with merge lot of issues introduced and << also there in file because of merge

PR #510, opened 2018-05-16
Issue #SB-1751 fix:notes

shakkeel1330:


src/app/client/src/app/modules/notes/components/note-card/note-card.component.ts
```diff
@@ -137,9 +137,11 @@ export class NoteCardComponent implements OnInit, OnChanges {
   }
 
   ngOnChanges() {
+    if (this.notesList) {
```
harishgilimi: This can be this.notesList = this.notesList || []

PR #540, opened 2018-05-21
Issue:feat:SB-3698 content player and collection player  for  unlisted. 

amantarento:


harishgilimi:
@amantarento  build is failing and collection player implementation is not correct

PR #637, opened 2018-06-07
Issue #SB-4304 fix: Library search filter

ArchanaKumari01:


src/app/client/src/app/modules/search/components/library-search/library-search.component.ts
```diff
@@ -219,11 +219,11 @@ export class LibrarySearchComponent implements OnInit {
         if (_.isEmpty(this.queryParams)) {
           this.filters = {};
         } else {
+          this.filters = {};
```
harishgilimi: Replace class property with local variable and pass as argument

PR #681, opened 2018-06-14
Issue #SB-3601 fix: Any qr scanner has to direct the user to the scanned content, App Play store link is not available on the page

prajwal88:


src/app/client/src/app/modules/public/public.module.ts
```diff
@@ -33,6 +35,6 @@ import { NgInviewModule } from 'angular-inport';
   ],
   declarations: [LandingPageComponent, SignupComponent, GetComponent, DialCodeComponent,
     PublicFooterComponent, PublicContentPlayerComponent, PublicCollectionPlayerComponent, ExploreContentComponent],
-  providers: [SignupService, PublicPlayerService, OrgManagementService]
+  providers: [SignupService, PublicPlayerService, OrgManagementService, DeviceDetectorModule, DeviceDetectorService]
```
anoophm: DeviceDetectorModule add it as import

PR #750, opened 2018-06-27
SB-4913 Sunbird portal cannot be deployed using the doc

rajeevsathish:


rajeevsathish:
will be creating a PR for 1.7.2

PR #803, opened 2018-07-12
Issue #SB-4175 fix:Library : Content search : The details of Grade in a content is not set properly.

ttplsudhir:
Before Fix 
grades
After Fix 
afterfux 1


ttplsudhir:
@lakhanmandloi - Can you give the some scenarios where this will fail if we not write css to not allow text wrap of each grade element. ?  For example ??

PR #812, opened 2018-07-13
Issue #SB-5388 fix: Ui bug fixes

prajwal88:


coolbung:
Please assign to @lakhanmandloi

coolbung:
Please assign to @lakhanmandloi

src/app/client/src/app/modules/core/components/main-header/main-header.component.html
```diff
@@ -1,16 +1,16 @@
 <div class="ui grid" *ngIf="!showExploreHeader">
-    <div class="center aligned twelve column row">
+    <div class="center aligned twelve column row" [class.header-height-before-login]="!userService.userid">
```
lakhanmandloi: need to add `middle` class here

src/app/client/src/app/modules/core/components/main-header/main-header.component.html
```diff
@@ -1,16 +1,16 @@
...
         <div class="one wide tablet computer only column"></div>
-        <div class="two wide tablet computer only column padding-top-15">
+        <div class="two wide tablet computer only column" [class.padding-top-5]="!userService.userid" [class.padding-top-15]="userService.userid">
```
lakhanmandloi: need to remove `padding-top-5` class from here.

src/app/client/src/app/modules/core/components/main-header/main-header.component.html
lakhanmandloi: `middle` class should be placed before `aligned` class.

src/app/client/src/app/modules/core/components/main-header/main-header.component.html
lakhanmandloi: Both suggestions are based on following - 

header has 80px height.. while logo has max 40px height.. padding top 5 will not align it in middle.

PR #821, opened 2018-07-18
Issue #SB-5575 fix: Validation for process ID field in check status component

vaishnavimanjunath:


src/app/client/src/app/modules/org-management/components/status/status.component.spec.ts
```diff
@@ -106,4 +106,17 @@ describe('StatusComponent', () => {
     component.ngOnDestroy();
     expect(component.unsubscribe$.complete).toHaveBeenCalled();
   });
+  it('should call getBulkUploadStatus method and pass else condition if processId contains only whiteSpaces', () => {
```
harishgilimi: write positive test case also when you pass proper process id it should call the service method

PR #828, opened 2018-07-19
Issue SB-5339 feat:Changed header as per design (instead of open filter)

ArchanaKumari01:


src/app/client/src/app/modules/core/components/prominent-filter/prominent-filter.component.ts
```diff
@@ -0,0 +1,255 @@
...
+                  const userRoles = formFieldCategory.allowedRoles.filter(element => this.userRoles.includes(element));
+                  if (!this.showField(formFieldCategory.allowedRoles)) {
+                    this.formFieldProperties.splice(this.formFieldProperties.indexOf(formFieldCategory), 1);
+                  }
+                }
+              });
+              this.getFormConfig();
+            },
+            (err: ServerResponse) => {
+              // this.toasterService.error(this.resourceService.messages.emsg.m0005);
```
anoophm: remove

PR #848, opened 2018-07-27
Issue #SB-5702 fix: removed 'fit' in the it statement of test case

revathitarento:


revathitarento:
Will be updated in other PR, as the change is very small

PR #902, opened 2018-08-17
Issue SB-5812 feat: Anonymous user - Content listed in the Library page should be relevant to the channel

ArchanaKumari01:


src/app/client/src/app/modules/public/module/explore/components/explore-content/explore-content.component.ts
```diff
@@ -223,14 +225,20 @@ export class ExploreContentComponent implements OnInit, OnDestroy {
             queryParams: this.queryParams
         });
     }
-
+    getFilters(filters) {
+        _.forEach(filters, (value) => {
+            if (value.code === 'board') {
+               this.dataDrivenFilter = _.get(value, 'range[0].name') ? _.get(value, 'range[0].name') : [];
```
anoophm: this.dataDrivenFilter. board = _.get(value, 'range[0].name') ? _.get(value, 'range[0].name') : [];

PR #904, opened 2018-08-17
SB-6387: Display tenant logo on public pages if configured

revathitarento:


src/app/client/src/app/app.component.spec.data.ts
```diff
@@ -1135,6 +1135,18 @@ export const mockData = {
...
             }
         }
+    },
+    tenantResponse: {
+        'id': 'api.tenant.info',
+        'params': {resmsgid: 'c33a73c0-a0b6-11e8-8b56-a93ad7c7024d', msgid: null, status: 'successful'},
+        'responseCode': 'OK',
+        'result': {
+            'titleName': 'rj',
+            'logo': 'https://staging.open-sunbird.org/assets/images/sunbird_logo.png',
```
rajeevsathish: multiple key has same value why can' t it be a like
x:
Y:
z: "ABC"

src/app/client/src/app/app.component.spec.ts
```diff
@@ -122,6 +122,20 @@ describe('AppComponent', () => {
     expect(document.title).toBe(mockData.tenantSuccess.result.titleName);
     expect(document.querySelector).toHaveBeenCalled();
   });
+  it('Should display the tenant logo if user is not logged in', () => {
+    const userService = TestBed.get(UserService);
```
rajeevsathish: why are you creating the const when you are not using it

src/app/client/src/styles.css
```diff
@@ -1145,7 +1145,12 @@ i.large.icon.limitedPublishingLinkIcon {
   width: auto;
   height: auto;
 }
-
+.tenant-header-logo-before-login {
```
rajeevsathish: is the Css approved by the UI team and the PM

src/app/client/src/app/app.component.spec.data.ts
revathitarento: updated different values for mockdata

src/app/client/src/app/app.component.spec.ts
revathitarento: removed the unused constant

src/app/client/src/styles.css
revathitarento: Css is updated as per the ui style guide, (Please check the attachment here - https://project-sunbird.atlassian.net/browse/SB-6387)

PR #939, opened 2018-09-03
Issue #SB-6895 fix:ORG Upload Instructions: Consistency should be there in uploads org's field/column name

ArchanaKumari01:
…ere in uploads org's field/column name


harishgilimi:
@ArchanaKumari01  PR is not covering all the scenarios what about **externalId** in sample csv and will the change in sample csv working where are test scenario's

PR #969, opened 2018-09-20
Sb 7135

revathitarento:


src/app/client/src/app/modules/core/components/main-header/main-header.component.ts
```diff
@@ -89,6 +88,7 @@ export class MainHeaderComponent implements OnInit, OnDestroy {
   */
   enableSignup = true;
   exploreRoutingUrl: string;
+  currentUrl: any;
```
rajeevsathish: why is this a type of any and not a type String?

src/app/client/src/app/modules/core/components/main-header/main-header.component.ts
```diff
@@ -100,9 +100,26 @@ export class MainHeaderComponent implements OnInit, OnDestroy {
...
     this.userService = userService;
     this.tenantService = tenantService;
-  }
+   }
 
   ngOnInit() {
+    this.router.events.pipe(filter(event => event instanceof NavigationEnd))
+      .subscribe(event => {
+        let currentRoute = this.activatedRoute.root;
+        while (currentRoute.children.length > 0) {
```
rajeevsathish: what happens if children is null?

revathitarento: currentRoute.children is an activatedRoute, so it will not be null.

src/app/client/src/app/modules/core/components/main-header/main-header.component.ts
revathitarento: changed to type string

src/app/client/src/app/modules/core/components/main-header/main-header.component.html
```diff
@@ -36,7 +36,7 @@
                     <a class="item" *ngIf='permissionService.permissionAvailable' appPermission [permission]='workSpaceRole' (click)="navigateToWorkspace()">{{resourceService.frmelmnts?.lnk?.workSpace}}</a>
                     <a class="item" *ngIf='permissionService.permissionAvailable' appPermission [permission]='myActivityRole' routerLink="myActivity">{{resourceService.frmelmnts?.lnk?.myactivity}}</a>
                     <a class="item" *ngIf='permissionService.permissionAvailable' appPermission [permission]='orgSetupRole' routerLink="orgType">{{resourceService.frmelmnts?.lbl?.orgtypes}}</a>
-                    <a class="item" (click)="logout()">{{resourceService.frmelmnts?.lnk?.logout}}</a>
+                    <a class="item" (click)="logout()" appTelemetryInteract [telemetryInteractEdata]="{ id: 'logout', type: 'click', pageid: currentUrl }">{{resourceService.frmelmnts?.lnk?.logout}}</a>
```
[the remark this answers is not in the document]
revathitarento: Forgot to update, updated now.

PR #1014, opened 2018-10-08
Issue #SB-6013 fix: Profile not working

ttplsudhir:


src/app/client/src/app/modules/core/services/user/user.service.ts
```diff
@@ -206,6 +206,9 @@ export class UserService {
     if (profileData.rootOrgId) {
       organisationIds.push(profileData.rootOrgId);
     }
+    if (profileData.skills === null) {
```
harishgilimi: This can be  `profileData.skills = _.get(profileData, 'skills' ) || []`

ttplsudhir: Requested changes Done.

PR #1051, opened 2018-10-22
Sb 8170 multiple config sources impl

loganathan1989:


src/app/helpers/config/configBuilder.js
```diff
@@ -0,0 +1,83 @@
+const _ = require('lodash')
+const cron = require('node-cron');
+let configSources = []
+let configKeyMap = {}
+let cacheRefreshEnabled = false
+const configHelper = require('./configHelper.js')
```
rajeevsathish: move all required at the top

PR #1099, opened 2018-11-14
Issue:SB-8137 User is forced to fill in board, medium and subject when they log in to Diksha

amantarento:


src/app/client/src/app/app.component.ts
```diff
@@ -21,6 +23,7 @@ import { Subscription } from 'rxjs';
   styleUrls: ['./app.component.css']
 })
 export class AppComponent implements OnInit {
+  @ViewChild('frameWorkPopUp') frameWorkPopUp;
```
harishgilimi: view child will be available after view init life cycle hook cross check once

src/app/client/src/app/modules/shared/interfaces/userProfile.ts
```diff
@@ -58,6 +58,7 @@ export interface IUserProfile {
     organisationIds?: Array<string>;
     hashTagIds?: Array<string>;
     organisationNames?: Array<string>;
+    framework?: any;
```
harishgilimi: What is the data type ? is it any ?

src/app/client/src/app/modules/resource/components/resource/resource.component.ts
```diff
@@ -77,25 +88,37 @@ export class ResourceComponent implements OnInit {
...
+      filters.subject = _.get(this.framework, 'subject');
+      filters.gradeLevel = _.get(this.framework, 'gradeLevel');
+      softConstraints = { subject: 98, medium: 99,  board: 100 };
+    }
     const option = {
       source: 'web',
       name: 'Resource',
-      filters: _.pickBy(this.filters, value => value.length > 0),
+      filters: filters,
+      softConstraints: softConstraints,
```
harishgilimi: This is ordering and sorting should be similar to explore page, contents with first board with index 0 or 1 should come first

PR #1118, opened 2018-11-19
Issue #000 fix: Card badge avatar alignment issue

lakhanmandloi:
**Before - Badge misaligned if data not available**

card-badge-before

**After - Badge aligned if data not available**
card-badge-after


lakhanmandloi:
@bvinayakumar Pls merge this pr.

PR #1168, opened 2018-11-30
Issue #000 : Fixes for Explore Course page

lakhanmandloi:
Before Fixes
explore-page-before

After Fixes
explore-course-after


lakhanmandloi:
@bvinayakumar Pls merge this pr. These were the changes requested by Namrita, Krishna and Nabeel in today's call hence there is no JIRA task for it. This Pr consists following -

1. Make blue, green  & Red colour consistent on all pages
2. In explore course page - Filter drop-down option text should be completely visible.
3. Each page section / Carousel heading should be in the small case except 1st letter should be capital.
4. Updated explore course page padding margin classes to new reusable format. This page was added later after we updated padding margin classes a few days ago, developer duplicated old page. 

Except explore course page, on all other pages, only colour changes have been made consistent hence screenshot not attached as its minor changes. For explore, page screenshot is already attached.

@harishgilimi - As per our today's call filter is made better now but it still has room for improvements. I was unable to complete the task as APIs were down and I was blocked. But still this PR can be merged as this improves the UI a bit more. Rest changes we can send in a separate PR.

PR #1180, opened 2018-12-06
Issue #SB-9321 fix: First card is not completely clickable

lakhanmandloi:


lakhanmandloi:
@bvinayakumar - not adding the screenshot bcoz this is not changing any visual element. A hidden element was preventing the click on the first card of the slider so that has been fixed in this.

PR #1192, opened 2018-12-10
Merging release 1.12 to master

harishgilimi:


G33tha:
@harishgilimi Can you please add the circleci config which is there in your master branch, here is the location of it in the sunbird ed "https://github.com/Sunbird-Ed/SunbirdEd-portal/tree/master/.circleci"

PR #1198, opened 2018-12-12
Issue#SB-8592 feat : Clicking on view all shows a filter selected to the user when one hasnt chosen a filter

amantarento:


src/app/client/src/app/modules/shared-feature/components/view-all/view-all.component.ts
```diff
@@ -247,14 +247,29 @@ export class ViewAllComponent implements OnInit, OnDestroy {
...
   private getContentList(request) {
+    let softConstraints = {};
+    softConstraints = _.get(this.activatedRoute.snapshot, 'data.softConstraints');
+    if (_.get(request, 'queryParams.softConstraintsFilter')) {
+      _.forIn(this.filters, (value, key) => {
+        if (_.has(softConstraints, key)) {
+          softConstraints =  _.omit(softConstraints, [key]);
+        }
+      });
+      const softConstraintsFilter = JSON.parse(request.queryParams.softConstraintsFilter);
```
harishgilimi: should handle error case if it is not an valid json string use try catch

PR #1279, opened 2018-12-31
Issue #000 fix: Remove lang dropdown in workspace and add missing text in hindi

lakhanmandloi:
Following things are in this pr -

1. Remove language switcher dropdown from the workspace area.
2. Added all language constants that are missing in Hindi. Due to the missing constant, some of the texts were not rendering.


lakhanmandloi:
@bvinayakumar - Pls merge this pr. Changes have been mentioned in the above comment.

PR #1333, opened 2019-01-16
Issue:SB-8498 [Consumption Channel] Depending on the language selected change the Interface Language

amantarento:


src/app/client/src/app/modules/shared/services/resource/resource.service.spec.data.ts
```diff
@@ -2498,5 +2498,53 @@ export const mockRes = {
...
+          },
+          'range': [
+           {
+            'value': 'en',
+            'name': 'English',
+            'dir': 'ltr'
+           },
+           {
+            'value': 'hi',
+            'name': 'हिंदी',
```
harishgilimi: This should be label

PR #1403, opened 2019-02-07
SB-9779 Users of generic portal should be able to choose from all boards in the portal filters

rajeevsathish:


rajeevsathish:
Closing as the implementation has to change to the creation of cache

PR #1408, opened 2019-02-07
Issue #SB-10244 feat: Search, discovery and enrolment into multiple open batches

anoophm:


harishgilimi:
@anoophm check the test cases which are failing

PR #1410, opened 2019-02-08
Issue #SB-10408 fix: copy content issue fix, test case fix version

raghuv45:


raghuv45:
> Pr description says test case fix, i dont see file changes in specfile.

test cases were failing randomly due to some code changes, which got fixed in this PR so it writen as test case fix.

PR #1411, opened 2019-02-08
Issue # SB-10307 fix: Telemetry missing for filter usage on Explore

raghuv45:


src/app/client/src/app/modules/core/components/data-driven-filter/data-driven-filter.component.ts
```diff
@@ -45,8 +45,10 @@ export class DataDrivenFilterComponent implements OnInit, OnChanges, OnDestroy {
   public isShowFilterPlaceholder = true;
 
   public filterIntractEdata: IInteractEventEdata;
```
anoophm: check the spelling

raghuv45: done

PR #1413, opened 2019-02-11
SB-9856 Removal of batch participants from invite-only batches

rajeevsathish:


harishgilimi:
@amantarento PR is merged which is having test cases

PR #1539, opened 2019-03-12
Issue #SB-10742 fix: lazy module for player in public module.

amantarento:


src/app/client/src/app/modules/public/public-routing.module.ts
```diff
@@ -71,7 +54,13 @@ const routes: Routes = [
   },
   {
     path: 'sign-in/sso', loadChildren: './module/sign-in/sso/sso.module#SsoModule'
-  }
+  },
+  {
+    path: 'play', loadChildren: './module/player/player.module#PlayerModule'
+  },
+  {
+    path: 'play', loadChildren: './module/player/player.module#PlayerModule'
```
anoophm: one route is enough

PR #1577, opened 2019-03-21
Release 1.15 10449

yravinderkumar33:


src/app/client/src/app/plugins/profile/components/update-user-details/update-user-details.component.ts
```diff
@@ -36,6 +38,24 @@ export class UpdateUserDetailsComponent implements OnInit, OnDestroy {
     this.getState();
   }
 
+  clearInput(e, element) {
+    const formControlName = element._element.nativeElement.attributes.getNamedItem('formcontrolname').value;
```
rajeevsathish: Can you please change the logic to get the Form control name by sending through the call

PR #1649, opened 2019-04-03
Release 1.15 batch list

yravinderkumar33:


src/app/client/src/app/modules/learn/services/course-batch/course-batch.service.ts
```diff
@@ -25,7 +25,8 @@ export class CourseBatchService {
         request: {
           filters: requestParam.filters,
           limit: requestParam.limit,
-          sort_by: requestParam.sort_by
+          sort_by: requestParam.sort_by,
```
bvinayakumar: For every updated ts file add / update corresponding unit tests.

PR #1711, opened 2019-04-15
Issue # SB-11272 fix: handling session expiry

anoophm:


anoophm:
> Test Cases are missing?

will be added when the feature is enabled. Currently its disabled

PR #1733, opened 2019-04-22
Issue #SB-11174 fix: telemetry impression oninit to onchanges

raghuv45:


src/app/client/src/app/modules/telemetry/directives/telemetry-impression/telemetry-impression.directive.ts
```diff
@@ -17,6 +17,8 @@ export class TelemetryImpressionDirective implements OnInit, OnDestroy {
    * reference of permissionService service.
   */
   public telemetryService: TelemetryService;
+  public eventTriggered = false;
```
rajeevsathish: why do we need Event Triggered here if this is true the impression will not be triggered right?

raghuv45: once Event Triggered becomes true impression event should not be generated, which is fine, and new instance of directive is initiated wherever impression directive is mentioend in the html pages, so no need to making it to false again.

PR #1890, opened 2019-05-16
added infinite scroll for cards list page

deveshMantra:


src/app/client/src/app/modules/dial-code-search/components/dial-code/dial-code.component.ts
```diff
@@ -22,6 +22,16 @@ export class DialCodeComponent implements OnInit, OnDestroy, AfterViewInit {
   /**
    * reference of SearchService
    */
+
+  /**
+   * Initializing the infinite scroller
+   */
+  itemsToDisplay: any = [];
+  initiallyItemsToLoad = 50;
```
rajeevsathish: why is the name initally Items to load if you are changing it in the method?

deveshMantra: the variable tells the initial items to be loaded when we first land to the page.
and since we need elements after once user scrolls page

PR #2036, opened 2019-05-25
Release 2.0.0 hotfix removing the CDN code from server.js and adding a gulp task

rajeevsathish:


rajeevsathish:
There will be a seperate PR to do the CDN Task

PR #2055, opened 2019-05-28
Issue SB-12709 Unable to view contents on click on 'view all' button …

rajeevsathish:
…in tenant specific explore page.


src/app/client/src/app/modules/public/module/explore/components/explore/explore.component.ts
```diff
@@ -92,7 +92,7 @@ export class ExploreComponent implements OnInit, OnDestroy, AfterViewInit {
       takeUntil(this.unsubscribe$))
       .subscribe((result) => {
         this.showLoader = true;
-        this.queryParams = { ...result[0], ...result[1] };
+        this.queryParams = { ...result[1] };
```
vinukumar-vs: If "slug" param is causing an issue, check for slug object to omit. Present code is blindly relaying on index, later if order changed then also it will fail.

PR #2091, opened 2019-06-03
Issue:SB-12022 Keep portal users on the portal when they click on the…

amantarento:
… SunbirdEd logo


src/app/client/src/app/modules/shared-feature/components/view-all/view-all.component.html
```diff
@@ -11,7 +11,7 @@
     <div class="sb-pageSection mt-24">
         <div class="sb-pageSection-header" *ngIf="!showLoader">
             <h4 class="sb-pageSection-title m-0 mr-5">{{sectionName}}</h4>
-            <span class="sb-pageSection-count sb-label sb-label-xs sb-label-error">{{totalCount}}</span>
+            <span  class="sb-pageSection-count sb-label sb-label-xs sb-label-error">{{totalCount}}</span>
```
rajeevsathish: why is there an extra space? And why you need to push this file as there is no changes

src/app/routes/clientRoutes.js
```diff
@@ -180,3 +180,27 @@ const loadTenantFromLocal = (req, res) => {
     renderDefaultIndexPage(req, res)
   }
 }
+const redirectTologgedInPage = (req, res) => {
+	let redirectRoutes = { '/explore': '/resources', '/explore/1': '/search/Library/1', '/explore-course': '/learn', '/explore-course/1': '/search/Courses/1' };
```
rajeevsathish: why is the first page only considered?
 and why are we hard coding the route /1???

PR #2101, opened 2019-06-03
IIssue #SB-10819 feat : File explorer opens up while uploading the content

Diptesh6501:


src/app/client/src/app/modules/offline/components/content-import/content-import.component.ts
```diff
@@ -64,6 +64,7 @@ export class ContentImportComponent implements OnInit, AfterViewInit {
       containerName: 'attachments/announcement',
     };
     this.offlineFileUploaderService.initilizeFileUploader(options);
+    $('input[type=file]').trigger('click');
```
harishgilimi: This doesn't works we need to go with id and input combination

PR #2172, opened 2019-06-14
Issue SB-0000 revert of using public service for api which dosnt requ…

rajeevsathish:
…ire auth token


rajeevsathish:
Issue #SB-0000 fix: using public service for api which dosnt require auth token #2158

PR #2565, opened 2019-09-05
SB-14150 - Support for new content types in textbook editor, collecti…

ArjunKhetia-NIIT:
SB-14150 - Support for new content types in textbook editor, collection editor, generic editor, and Workspace


src/app/client/src/app/modules/search/components/user-filter/user-filter.component.ts
```diff
@@ -66,7 +66,7 @@ export class UserFilterComponent implements OnInit {
       });
   }
 
-  private subscribeToQueryParams() {
+  public subscribeToQueryParams() {
```
rajeevsathish: this should be private and just do a tobe called

PR #2669, opened 2019-10-01
Issue #SB-13991 feat: added content update

swetha-aggidevara:


src/app/client/src/app/modules/offline/services/download-manager/download-manager.service.ts
```diff
@@ -53,4 +54,31 @@ export class DownloadManagerService {
...
+      data: {
+        request: {
+          parentId:  _.get(data, 'parentId')
+        }
+      }
+    };
+  return this.publicDataService.post(requestParams).pipe(
+    map((result) => {
+      this.toasterService.info('Updating...');
+      this.downloadEvent.emit('Update started');
```
harishgilimi: event Emiiter name should not contain space it should be camelCase

src/app/client/src/app/modules/offline/services/download-manager/download-manager.service.ts
```diff
@@ -53,4 +54,31 @@ export class DownloadManagerService {
...
+      return result;
+    }),
+    catchError((err) => {
+      return observableThrowError(err);
+    }));
+  }
+
+  getContent(content) {
```
harishgilimi: This is duplicate code

PR #2701, opened 2019-10-18
Telemetry sb 14956

venkanagouda-k:
https://project-sunbird.atlassian.net/browse/SB-14956


src/app/client/src/app/modules/cbse-program/components/chapter-list/chapter-list.component.ts
```diff
@@ -39,7 +41,7 @@ export class ChapterListComponent implements OnInit, OnChanges {
   public question_categories: any;
   public question_type: Array<any> = [];
```
venkateshwarans: Can we name them differently.?  We already have variables questionCategories and questionType

PR #2735, opened 2019-10-24
Issue #SB-1321 fix:District mapping updates

deveshMantra:


src/app/client/src/app/modules/core/services/device-register/device-register.service.ts
```diff
@@ -40,6 +43,28 @@ export class DeviceRegisterService  {
...
+  fetchDeviceProfile() {
+    const httpOptions: HttpOptions = {
+      headers: {
+        'Content-Type': 'application/json'
+      }
+    };
+    const systemSetting = {
+      url: this.configService.urlConFig.URLS.SYSTEM_SETTING.CUSTODIAN_ORG,
+    };
+    return this.http.get('https://dev.sunbirded.org/api/v3/device/profile/' + this.deviceId, httpOptions);
```
rajeevsathish: Need to change the get path

PR #2796, opened 2019-11-11
ssue #SB-15405 fix: User sees the message as wrong from the download information popup

swetha-aggidevara:
… information popup


src/app/client/src/app/modules/shared/components/offline-card/offline-card.component.ts
```diff
@@ -70,12 +71,16 @@ export class OfflineCardComponent implements OnInit, OnChanges, OnDestroy {
   }
 
   public onAction(data, action) {
+    console.log('dataaa', data);
```
harishgilimi: remove console logs

src/app/client/src/app/modules/player-helper/components/content-download/content-download.component.ts
```diff
@@ -107,9 +108,13 @@ export class ContentDownloadComponent implements OnInit, OnDestroy {
   }
 
   download(content) {
+    console.log('content', content);
```
harishgilimi: remove console log

PR #2835, opened 2019-11-18
Issue#SB-15803 feat: added mising successfull icon and some validations chnages in report other issue components and created service files in services for report other issue

santoshilimi:


harishgilimi:
Rejected Due to number of files

PR #2843, opened 2019-11-19
Issue #SC-1473 fix:migration of keycloack from 3 to 6

deveshMantra:


src/app/package.json
```diff
@@ -94,7 +94,7 @@
     "js-md5": "^0.5.0",
     "jsonwebtoken": "^7.4.1",
     "kafka-node": "^4.1.1",
-    "keycloak-connect": "git+https://github.com/project-sunbird/keycloak-nodejs-connect.git#release-2.3.5",
+    "keycloak-connect": "git+https://github.com/project-sunbird/keycloak-nodejs-connect.git#release-2.5.0",
```
vinukumar-vs: Create this as an NPM package and publish it. Use the publishing version of the NPM package here instead of depending on the git branch,

@rajeevsathish

PR #2850, opened 2019-11-20
Issue #SB-15678 feat : Handling User Profile on Portal when Tenant's …

yravinderkumar33:
…Framework doesn't have a Board

### DIFF
--- a/src/app/client/src/app/modules/shared-feature/components/profile-framework-popup/profile-framework-popup.component.ts
+++ b/src/app/client/src/app/modules/shared-feature/components/profile-framework-popup/profile-framework-popup.component.ts
@@ -205,13 +205,16 @@ export class ProfileFrameworkPopupComponent implements OnInit, OnDestroy {
   }
   onSubmitForm() {
     const selectedOption = _.cloneDeep(this.selectedOption);
-    selectedOption.board = [this.selectedOption.board];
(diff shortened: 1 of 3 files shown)


src/app/client/src/app/modules/shared-feature/components/profile-framework-popup/profile-framework-popup.component.ts
```diff
@@ -205,13 +205,16 @@ export class ProfileFrameworkPopupComponent implements OnInit, OnDestroy {
...
     const selectedOption = _.cloneDeep(this.selectedOption);
-    selectedOption.board = [this.selectedOption.board];
+    selectedOption.board = _.get(this.selectedOption, 'board') ? [this.selectedOption.board] : null;
     selectedOption.id = this.frameWorkId;
     this.submit.emit(selectedOption);
   }
   private enableSubmitButton() {
-    if (_.get(this.selectedOption, 'board.length') && _.get(this.selectedOption, 'medium.length')
-      && _.get(this.selectedOption, 'gradeLevel.length')) {
+    const optionalFields = ['subject'];
```
vinukumar-vs: Why only "subject" field is optional?

yravinderkumar33: Because it's an optional field in the form while board, medium and subject fields are mandatory .
If form doesn't comes in the framework then subject and medium becomes the mandatory fields.

PR #2878, opened 2019-11-25
Issue #SB-15822 feat: User preference data saving implementation and test cases

souravdey091:


src/app/client/projects/desktop/src/app/modules/offline/components/onboarding-user-preference/onboarding-user-preference.component.spec.data.ts
```diff
@@ -0,0 +1,367 @@
+export const onboarding_user_preference_test = {
```
harishgilimi: Update test data

PR #2928, opened 2019-12-02
Issue #SB-15659 bug:added missing successful message toaster in copy pendrive option

santoshilimi:
… pendrive feature


src/app/resourcebundles/data/consumption/en.properties
```diff
@@ -806,6 +806,7 @@ frmelmnts.lbl.desktopAppFeature001 = Free unlimited content
 frmelmnts.lbl.desktopAppFeature002 = Multilingual support
 frmelmnts.lbl.desktopAppFeature003 = Play content offline
 messages.fmsg.m0090 = Could not download. Try again later
+messages.fmsg.m0098 = Imported completed.
```
anoophm: are we adding a success message for import or export? A success message should be added in messages.smsg not in messages.fmsg

PR #3021, opened 2019-12-16
Issue#SB-14670 fix Show derived source information on the Portal when…

ArjunKhetia-NIIT:
… a course is copied


src/app/client/src/app/modules/player-helper/components/content-credits/content-credits.component.ts
```diff
@@ -35,12 +36,18 @@ export class ContentCreditsComponent implements OnInit, OnChanges {
...
+showCopiedContent = false;
+ instance: string;
 
   constructor(resourceService: ResourceService) {
     this.resourceService = resourceService;
   }
 
   ngOnInit() {
+    this.instance = _.upperCase(this.resourceService.instance);
+    if (window.location.href.toString().includes('course')) {
```
rajeevsathish: This is not the correct way to check if this is a course or not please check if we have an alternative

PR #3024, opened 2019-12-16
Issue #SB-16372 feat: Test cases for update user location and content preferences

santoshilimi:


santoshilimi:
duplicate due to same code is pushed to different PR

PR #3028, opened 2019-12-17
Issue#SB-14670 fix Show derived source information on the Portal when…

ArjunKhetia-NIIT:
… a course is copied


src/app/resourcebundles/data/consumption/en.properties
```diff
@@ -223,6 +223,8 @@ frmelmnts.lbl.anncmnttblseen = Seen
 frmelmnts.lbl.anncmnttblsent = Sent
 frmelmnts.lbl.attributions = Attributions
 frmelmnts.lbl.author = Author
+frmelmnts.lbl.authorofsourcecontent = Author of Source Content
+frmelmnts.lbl.contentderivedfrom = This Content is Derived From
```
rajeevsathish: frmelmnts.lbl.contentcopiedtitle = This content is derived from 
this is already available so why do we need one more

ArjunKhetia-NIIT: Removed.

PR #3109, opened 2019-12-26
Issue #SB-16843: bug : Desktop app: User is not able to close the player window, after he opens multiple contents from Content manager.

santoshilimi:
…yer window, after he opens multiple contents from Content manager.


harishgilimi:
Rejected since the implementation is not correct

PR #3127, opened 2019-12-30
Issue#SB-16901: Location: post entering the state values users are unable to select the state and district values

ArjunKhetia-NIIT:


vinukumar-vs:
@ArjunKhetia-NIIT No unit-testcases & No JIRA issue tagged to it.
@rajeevsathish

PR #3129, opened 2019-12-30
Issue #SB-16901 Location post entering the state values users are unable to select the state and district values

ArjunKhetia-NIIT:


src/app/client/src/app/modules/shared-feature/components/user-location/user-location.component.ts
```diff
@@ -259,21 +259,26 @@ export class UserLocationComponent implements OnInit {
...
+          if (this.districtDiv.query && event.target.value) {
+            this.userDetailsForm.get(formControlName).reset();
+            this.districtDiv.query = '';
+            this.districtDiv.focus();
+          }
           break;
         }
       }
-    }
+    }, 500);
```
vinukumar-vs: Avoid timeout if we dont know the exact time. This may not work in all cases, especially in low configuration devices(desktops/laptops)

PR #3152, opened 2020-01-05
Issue #SB-16106 feat: Created a credits-and-licence component

itsvick:


src/app/client/projects/desktop/src/app/modules/content-consumption/components/credits-and-licence/credits-and-licence.component.ts
```diff
@@ -0,0 +1,22 @@
...
+})
+export class CreditsAndLicenceComponent {
+
+  @Input() contentData;
+  constructor(public resourceService: ResourceService) { }
+
+  embedUrl(text: string) {
+    if (text) {
+      const urlRegex = /(https?:\/\/[^\s]+)/g;
+      return text.replace(urlRegex, '<a href="$1">$1</a>');
```
harishgilimi: What happens when we click on the url ?

itsvick: I missed that scenario, but I think we should open it in the browser.

PR #3166, opened 2020-01-08
Issue#SB-16558 UI support - Desktop app - UI fixes

ArtiSingh23:


src/app/client/src/assets/styles/components/_video.scss
```diff
@@ -65,6 +65,7 @@
     video {
       width: 100%;
       height: 100%;
+      border: 6px solid var(--black);
```
sri-pusuluri: px has to be in REM.

PR #3233, opened 2020-01-13
issue #SB-17171 feat : test cases for content player component

santoshilimi:


src/app/client/projects/desktop/src/app/modules/content-consumption/components/content-player/content-player.component.ts
```diff
@@ -141,7 +139,7 @@ export class ContentPlayerComponent implements AfterViewInit, OnChanges {
   }
 
   generateScoreSubmitEvent(event: any) {
-    if (event.data.toLowerCase() === (this.CONSTANT.ACCESSEVENT).toLowerCase()) {
+    if (_.lowerCase(event.data) === (_.lowerCase(this.CONSTANT.ACCESSEVENT))) {
```
harishgilimi: should  use _.toLower instead of  _.lowerCase

PR #3288, opened 2020-01-17
Issue #SB-15452 fix : Integrate resource preview and CKEditor math plugin 

vaibhavbhuva:


src/app/helpers/latexService.js
```diff
@@ -45,6 +46,11 @@ function svg2png(svgString) {
     });
 }
 
+function png2base64(pngString) {
```
cksachdev: @vrayulu We have added this method to convert the formula to base64. 
cc: @vinukumar-vs

src/app/client/src/assets/libs/mathEquation/plugin/mathTextPlugin.js
```diff
@@ -0,0 +1,126 @@
...
+            this.listenTo(view, 'execute', () => {
+                this._loadIframeModal('btnClick');
+            });
+            return view;
+        });
+
+        this.listenTo( editor.editing.view.document, 'click', ( evt, data ) => {
+            if ( data.domEvent.detail == 3 ) {
+                let ifrm = document.getElementById('mathModalIframe')
+                if (data.domTarget.nodeName.toLowerCase() == 'img' && data.domTarget.hasAttribute("data-mathtext") && !ifrm) {
```
cksachdev: @vinukumar-vs data-mathtext attribute is added to store the formula. This will help for the migration of existing questions. 
cc: @pallakartheekreddy @Nik720 @vaibhavbhuva

PR #3290, opened 2020-01-17
Issue #SB-15452 fix: fixing appIcons for resource templates

venkateshwarans:


src/app/client/src/app/modules/program/components/program/data.ts
```diff
@@ -1116,8 +1116,7 @@ const programThirdSession = {
                     'audience': [
                       'Learner'
                     ],
-                    // tslint:disable-next-line:max-line-length
-                    'appIcon': 'https://ntpstagingall.blob.core.windows.net/ntp-content-staging/content/do_21291553051403878414/artifact/explanation.thumb_1576602846206.png',
+                    'appIcon': 'https://ntpproductionall.blob.core.windows.net/ntp-content-production/content/do_31280847585016217619710/artifact/explanation.thumb.png',
```
cksachdev: This could be a configurable asset path based on environment. Possible tech backlog for next release.
cc: @vinukumar-vs @pallakartheekreddy

pallakartheekreddy: @cksachdev This file is just for reference. When someone is creating a program, they will give the path based on the env.
cc: @vinukumar-vs

PR #3325, opened 2020-01-21
Issue #SB-17365 fix: Integrated telemetry in the question-list and CKEditor math plugin issue fixes

vaibhavbhuva:


src/app/client/src/app/modules/cbse-program/components/question-list/question-list.component.ts
```diff
@@ -59,25 +65,29 @@ export class QuestionListComponent implements OnInit, OnDestroy {
   public goToNextQuestionStatus = true;
   previewBtnVisibility = true;
   visibility: any;
-  @ViewChild('resourceTtlTextarea') resourceTtlTextarea: ElementRef;
+  telemetryImpression: any;
+  public telemetryPageId = 'question-list';
 
   constructor(
-    private configService: ConfigService, private userService: UserService,
+    public configService: ConfigService, public userService: UserService,
```
cksachdev: @vaibhavbhuva What is the reason to make service as public?

vaibhavbhuva: @cksachdev We are using this service in the DOM for telemetry. So for that, we have marked it as public.

PR #3329, opened 2020-01-21
Issue #SB-15452 fix: program components as a service to load dynamically

venkateshwarans:


venkateshwarans:
It is no longer required.

PR #3334, opened 2020-01-21
Issue #SB-15452 fix: program components as a service to load dynamically

pallakartheekreddy:


src/app/client/src/app/modules/program/components/program/program.component.ts
```diff
@@ -207,7 +207,7 @@ export class ProgramComponent implements OnInit, OnDestroy, AfterViewInit {
       if (this.tabs && this.programDetails.userDetails) {
         this.defaultView = _.find(this.tabs, { 'index': this.getDefaultActiveTab() });
         this.programStageService.addStage(this.defaultView.onClick);
-        this.component = this.componentMapping[this.defaultView.onClick];
+        this.component = this.programComponentsService.getComponentInstance(this.defaultView.onClick);
```
vinukumar-vs: Remove the componentMapping object if we are not using, which is defined as public.

src/app/client/src/app/modules/program/services/program-components/program-components.service.spec.ts
```diff
@@ -0,0 +1,12 @@
+import { TestBed } from '@angular/core/testing';
+
+import { ProgramComponentsService } from './program-components.service';
+
+describe('ProgramComponentsService', () => {
+  beforeEach(() => TestBed.configureTestingModule({}));
+
+  xit('should be created', () => {
```
vinukumar-vs: There is no test-cases for this component service.

PR #3383, opened 2020-01-27
Issue #SB-17517 fix: dial code is fixed

swetha-aggidevara:
…t inside it


src/app/client/projects/desktop/src/app/modules/offline/components/content-manager/content-manager.component.html
```diff
@@ -1,5 +1,5 @@
 <sui-accordion class="styled fluid sb-single-pannel-accordion sb-content-manager-panel"
-  [ngClass]="{'sb-single-pannel-accordion-fixed': isOpen === true}">
+  [ngClass]="{'sb-single-pannel-accordion-fixed': isOpen === true}" *ngIf="showDownloadManager">
```
anoophm: can u use contentList.length instead

PR #3384, opened 2020-01-27
Issue #SB-17500 fix: content-manager is disabled if there is no conten…

swetha-aggidevara:
…t inside it


rajeevsathish:
Merging as approved by @sharathkashyap

PR #3388, opened 2020-01-27
Issue #SB-17522 fix: Showing default framework as selected values in My Download Section when framework is unavailable locally

itsvick:
…My Downloads filters when framework is unavailable


src/app/client/projects/desktop/src/app/modules/offline/components/library-filters/library-filters.component.ts
```diff
@@ -57,6 +58,7 @@ export class LibraryFiltersComponent implements OnInit, OnDestroy {
   ) { }
 
   ngOnInit() {
+    this.showDefaultFilter = false;
```
anoophm: why is this flag initialized twice with the same value

rajeevsathish:
Merging as approved by @sharathkashyap

PR #3393, opened 2020-01-27
Issue #SB-17531 fix: recent downloaded data is coming in content man…

swetha-aggidevara:
…ager


src/app/client/projects/desktop/src/app/modules/offline/components/content-manager/content-manager.component.ts
```diff
@@ -25,6 +25,7 @@ export class ContentManagerComponent implements OnInit, OnDestroy {
   apiCallTimer = timer(1000, 3000).pipe(filter(data => !data || (this.callContentList)));
   apiCallSubject = new Subject();
   completedCount: number;
+  localContentData: any = [];
```
anoophm: no need to create new property

PR #3395, opened 2020-01-27
Issue #SB-16977 fix: Portal back-end test cases

sudip-mukherjee:


src/app/constants/routes.constants.js
```diff
@@ -0,0 +1,8 @@
+const PORTAL_PORT = require('./../helpers/environmentVariablesHelper').PORTAL_PORT
```
rajeevsathish: please rename the file as routesConstants as followed in all places

PR #3403, opened 2020-01-28
Merging Release-2.7.0 into release-2.8.0

rajeevsathish:


rajeevsathish:
As discussed with @gandham-santhosh merging this PR but the coverage% is less.as program-portal is merged into portal codebase in release-2.7.0 and there is no test case for programs-portal

PR #3469, opened 2020-01-31
Issue  #SB-15452 fix: Integrated test-cases and telemetry VSA , LA, SA and MCQ

vaibhavbhuva:


rajeevsathish:
As Discussed with @vinukumar-vs and @gandham-santhosh the coverage will improve in the coming week so merging this PR

PR #3518, opened 2020-02-06
Issue #SB-17729 bug: User should be able to get the same popup after clicks on "Load content" from the top right corner of the My downloads page

santoshilimi:


santoshilimi:
change in version

santoshilimi:
close

PR #3524, opened 2020-02-06
Issue #SB-16917 fix: Post signup content starts to play in the backgr…

ArjunKhetia-NIIT:
…ound before completing the onboarding details

### DIFF
--- a/src/app/client/src/app/app.component.ts
+++ b/src/app/client/src/app/app.component.ts
@@ -17,6 +17,7 @@ import { first, filter, mergeMap, tap, map, skipWhile, startWith, takeUntil } fr
 import { CacheService } from 'ng2-cache-service';
 import { DOCUMENT } from '@angular/platform-browser';
 import { ShepherdService } from 'angular-shepherd';
+import { PopupControlService } from './service/popup-control.service';
 
 /**
  * main app component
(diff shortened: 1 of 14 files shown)


src/app/client/src/app/service/popup-control.service.ts
```diff
@@ -0,0 +1,58 @@
...
+  public showUserVerificationPopup = false;
+  public isLocationConfirmed = true;
+
+  public popupSource = new BehaviorSubject(null);
+  checkPopupStatus = this.popupSource.asObservable();
+
+  constructor() { }
+
+  changePopupStatus(modalkey: string, modalvalue: boolean) {
+    if (modalkey === 'showFrameWorkPopUp') {
```
rajeevsathish: why cannot this be a switch

src/app/client/src/app/app.component.ts
```diff
@@ -195,10 +196,12 @@ export class AppComponent implements OnInit, OnDestroy {
           if (!deviceProfile.userDeclaredLocation ||
             !(this.usersProfile && this.usersProfile.userLocations && this.usersProfile.userLocations.length >= 1)) {
             this.isLocationConfirmed = false;
+            this.popupControlService.changePopupStatus('isLocationConfirmed', false);
```
rajeevsathish: the if and else are doing the same call why do you want to call it 2times

src/app/client/src/app/service/popup-control.service.ts
ArjunKhetia-NIIT: Done..

src/app/client/src/app/app.component.ts
ArjunKhetia-NIIT: Sir That If-Else is already been written. I have No idea about it.

PR #3584, opened 2020-02-24
fix: reorder of package install

keshavprasadms:


vinukumar-vs:
This is not required. Merging R2.7 into R 2.8 will solve this issue.
https://github.com/Sunbird-Ed/SunbirdEd-portal/pull/3583

PR #3589, opened 2020-02-25
Issue #SB-17859 fix: [Telemetry] Tags value should not display multip…

yravinderkumar33:
…le times.


src/app/client/src/app/modules/workspace/components/content-editors/collection-editor/collection-editor.component.ts
```diff
@@ -205,7 +205,7 @@ export class CollectionEditorComponent implements OnInit, OnDestroy {
         type: 'User'
       },
       contextRollUp: this.telemetryService.getRollUpData(this.userProfile.organisationIds),
-      tags: this.userService.dims,
+      tags: _.uniq(this.userService.dims),
```
vinukumar-vs: Why usreService.dims having duplicates?

PR #3595, opened 2020-02-25
Issue #SB-0000 fix : reports file path changes for manage user reports.

yravinderkumar33:


src/app/client/src/app/modules/manage/components/user-org-management/user-org-management.component.ts
```diff
@@ -160,7 +160,7 @@ export class UserOrgManagementComponent implements OnInit, AfterViewInit {
   }
 
   public getUserJSON() {
-    this.manageService.getData(this.slug, this.userJSON).subscribe(
+    this.manageService.getData(this.userJSON, `${this.slug}.json`).subscribe(
```
vinukumar-vs: Why don't you create the map & use it instead of so many variables & functions defined for this?

src/app/routes/reportRoutes.js
```diff
@@ -16,7 +16,8 @@ module.exports = function (app) {
 
     app.get('/admin-reports/:slug/:filename',
         proxyUtils.verifyToken(),
-        reportHelper.validateSlug(),
+        reportHelper.validateSlug(['geo-summary', 'geo-detail', 'geo-summary-district', 'user-summary', 'user-detail',
```
vinukumar-vs: You can use the map object here as well.

src/app/client/src/app/modules/manage/components/user-org-management/user-org-management.component.ts
yravinderkumar33: Yes, I agree.  This component needs refactoring because of duplicate code. Due to time constraints, I made the changes in the existing code itself.

PR #3611, opened 2020-02-28
Issue #SB-18013 fix: Content type Field is not displayed for Resource and upload content in Up for review section

sudip-mukherjee:


src/app/client/src/app/modules/player-helper/components/content-player-metadata/content-player-metadata.component.html
```diff
@@ -85,7 +85,7 @@ <h4 class="sb-section-title text-capitalize mb-16 d-block">
         <dd *ngIf="metadata.keywords">{{metadata.keywords}}</dd>
     </dl>
     <dl class="sb-meta-data">
-        <dt>{{resourceService.frmelmnts.lbl.resourcetype}}</dt>
+        <dt>{{resourceService.frmelmnts.lbl.contentType}}</dt>
```
vinukumar-vs: This has changed to ContentType but the below line is pointing to resourceType.

sudip-mukherjee: We are showing the resourceType only (getting from metadata), just the label has to change from Resource Type to Content Type.

PR #3662, opened 2020-03-10
Issue #SB-18213 Teachers data: Admin is getting View more and downloa…

rajeevsathish:
…d button even if there is not data available


rajeshkumaravel:
- Verified

PR #3707, opened 2020-03-16
Cleanup

rajeshkumaravel:
- Replaced variable with instance


src/app/client/src/app/modules/shared/components/offline-application-download/offline-application-download.component.html
```diff
@@ -31,7 +31,7 @@ <h2 class="offline-download-card-heading">
     <div class="five wide column pl-0">
       <div class="offline-download-card offline-page-content">
         <div class="d-flex flex-ai-center flex-jc-center text-center flex-dc offline-page-content">
-          <img src="../assets/images/offline-computer.png" alt="download diksha app" class="desktop-icon">
+          <img src="../assets/images/offline-computer.png" alt="download {{instance}} app" class="desktop-icon">
```
rajeevsathish: if you use instance the value will be DIKSHA, not diksha is it ok

rajeshkumaravel: Yes; since it is a label (alt for img) does not have an impact on functionality.

PR #3771, opened 2020-03-26
Issue #SB-18083 fix: User is not returning to corresponding pages from where they have initiated the QR code search

JagadishPujari:
[JIRA LINK](https://project-sunbird.atlassian.net/browse/SB-18083)


JagadishPujari:
The approach is not correct so closing PR

PR #3774, opened 2020-03-27
Issue #SB-18310 fix: Sorting subjects and books in ascending order

rajeshkumaravel:
- Issue #SB-18310 fix: Sorting subjects and books in ascending order


rajeshkumaravel:
@rajeevsathish Please review.

vinukumar-vs:
@rajeshkumaravel  Why are we not using angular pipes?
Example: https://www.npmjs.com/package/ngp-sort-pipe

@rajeevsathish @yravinderkumar33 @deveshMantra

rajeshkumaravel:
> @rajeshkumaravel Why are we not using angular pipes?
> Example: https://www.npmjs.com/package/ngp-sort-pipe
> 
> @rajeevsathish @yravinderkumar33 @deveshMantra

@vinukumar-vs 
- Plugin can be used
- The top level sort i.e Subject name can be sorted using plugin
- However at this https://github.com/Sunbird-Ed/SunbirdEd-portal/blob/2e071075e533edefb118248b263baefaeca80c64/src/app/client/src/app/modules/public/module/explore/components/explore/explore.component.html#L8

We need to pass the sorted books list.

PR #3778, opened 2020-03-30
Issue #SB-18083 fix: User is not returning to corresponding pages from where they have initiated the QR code search

JagadishPujari:
[JIRA LINK](https://project-sunbird.atlassian.net/browse/SB-18083)

### DIFF
--- a/src/app/client/src/app/modules/public/module/player/components/public-content-player/public-content-player.component.ts
+++ b/src/app/client/src/app/modules/public/module/player/components/public-content-player/public-content-player.component.ts
@@ -147,28 +147,16 @@ export class PublicContentPlayerComponent implements OnInit, OnDestroy, AfterVie
    * closes conent player and revert to previous url
    * @memberof ContentPlayerComponent
    */
-  close() {
-    try {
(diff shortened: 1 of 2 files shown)


src/app/client/src/app/modules/public/module/player/components/public-content-player/public-content-player.component.ts
```diff
@@ -147,28 +147,16 @@ export class PublicContentPlayerComponent implements OnInit, OnDestroy, AfterVie
...
-        } else if (this.isOffline) {
-            this.navigationHelperService.navigateToResource('');
-          } else {
-            this.navigationHelperService.navigateToResource('/explore');
-          }
-      }, 100);
-    }
+  public close() {
+    let previousUrl = _.get(this.navigationHelperService.getPreviousUrl(), 'url');
+      if (_.includes(previousUrl, 'play')) {
```
vinukumar-vs: This logic is taking user always to resource/explore page. Ideally it should take the user from where he came. It should be like browser back button behaviour.

@anoophm

PR #3795, opened 2020-04-02
Issue #SB-16280 fix: [Workspace] Incorrect message get displayed if there are no records in all content bucket

JagadishPujari:
[JIRA LINK](https://project-sunbird.atlassian.net/browse/SB-16280)

### DIFF
--- a/src/app/resourcebundles/data/consumption/en.properties
+++ b/src/app/resourcebundles/data/consumption/en.properties
@@ -879,6 +879,7 @@ messages.stmsg.m0140 = DOWNLOADING
 messages.stmsg.m0141 = Data unavailable to generate Score Report
 messages.stmsg.m0142 = PAUSED
 messages.stmsg.m0143 = DOWNLOAD
+messages.stmsg.m0144 = No Content Available
 orgname = orgname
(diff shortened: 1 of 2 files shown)


src/app/resourcebundles/data/consumption/en.properties
```diff
@@ -879,6 +879,7 @@ messages.stmsg.m0140 = DOWNLOADING
 messages.stmsg.m0141 = Data unavailable to generate Score Report
 messages.stmsg.m0142 = PAUSED
 messages.stmsg.m0143 = DOWNLOAD
+messages.stmsg.m0144 = No Content Available
```
anoophm, line 882: check this message already exist or not. If exist use that

src/app/resourcebundles/data/consumption/en.properties
```diff
@@ -879,6 +879,7 @@ messages.stmsg.m0140 = DOWNLOADING
...
 orgname = orgname
 participants = Participants
```
anoophm, line 884: add en.json file as well. These messages needs to be added in phaser app as well.

PR #3812, opened 2020-04-03
Issue #SB-18664 feat: Assess event sending to server should happen fo…

vinukumar-vs:
Assessment sending to server should work for content type onboardingresource as well

### DIFF
--- a/src/app/client/src/app/modules/learn/services/assessment/assessment-score.service.ts
+++ b/src/app/client/src/app/modules/learn/services/assessment/assessment-score.service.ts
@@ -79,7 +79,7 @@ export class AssessmentScoreService {
    */
   private checkContentForAssessment() {
     if (_.get(this._batchDetails, 'batchId') && _.get(this._contentDetails, 'identifier') && _.get(this._batchDetails, 'courseId')) {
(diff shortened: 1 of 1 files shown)


src/app/client/src/app/modules/learn/services/assessment/assessment-score.service.ts
```diff
@@ -79,11 +79,9 @@ export class AssessmentScoreService {
    */
   private checkContentForAssessment() {
     if (_.get(this._batchDetails, 'batchId') && _.get(this._contentDetails, 'identifier') && _.get(this._batchDetails, 'courseId')) {
-      if (this._contentDetails && _.get(this._contentDetails, 'contentType') === 'SelfAssess') {
-        this.initialized = true;
-      } else {
-        this.initialized = false;
-      }
+      this.initialized = false;
```
rajeevsathish, line 82: why do we need an if and else condition if we do that same in both the case?

vinukumar-vs: It was wrong check, I have made changes.

PR #3816, opened 2020-04-04
Issue #SB-0000 fix:adding organization id for pageassemble api

deveshMantra:


src/app/client/src/app/modules/public/module/course/components/public-course/public-course.component.ts
```diff
@@ -132,6 +132,7 @@ export class PublicCourseComponent implements OnInit, OnDestroy, AfterViewInit {
     const option = {
       source: 'web',
       name: 'AnonymousCourse',
+      organisationId: this.hashTagId || '*',
```
vinukumar-vs, line 135: is this.hashTagId or  this.userService.hashTagId,

In the leaner-page.component.ts you are using 
 this.userService.hashTagId

deveshMantra: it is non-logged user. hence userserive.hastagid will not be present.

PR #3831, opened 2020-04-07
Issue #SB-18417 fix: Change the label for QR code size

pradoshKumar:


src/app/client/src/assets/styles/components/_certificate-trained.scss
```diff
@@ -80,6 +84,16 @@
   }
 }
 
+html[dir="rtl"] {
```
sri-pusuluri: try to mention text-left, this will be taken care of rtl

PR #3832, opened 2020-04-07
Issue #SB-17311 fix: District list is alphabetical order AND District name starting with Caps letter

JagadishPujari:
[JIRA LINK](https://project-sunbird.atlassian.net/browse/SB-17311)
[JIRA LINK](https://project-sunbird.atlassian.net/browse/SB-16203)

### DIFF
--- a/src/app/client/src/app/modules/dashboard/components/data-chart/data-chart.component.ts
+++ b/src/app/client/src/app/modules/dashboard/components/data-chart/data-chart.component.ts
@@ -109,7 +109,7 @@ export class DataChartComponent implements OnInit, OnDestroy {
           this.dateFilterReferenceName = filter.reference;
         }
         this.filtersFormGroup.addControl(_.get(filter, 'reference'), this.fb.control(''));
-        filter.options = _.uniq(_.map(this.chartData, data => data[filter.reference].toLowerCase()));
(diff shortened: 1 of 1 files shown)


src/app/client/src/app/modules/dashboard/components/data-chart/data-chart.component.ts
```diff
@@ -109,7 +109,7 @@ export class DataChartComponent implements OnInit, OnDestroy {
           this.dateFilterReferenceName = filter.reference;
         }
         this.filtersFormGroup.addControl(_.get(filter, 'reference'), this.fb.control(''));
-        filter.options = _.uniq(_.map(this.chartData, data => data[filter.reference].toLowerCase()));
+        filter.options = _.sortBy(_.uniq(_.map(this.chartData, data => _.capitalize(data[filter.reference].toLowerCase()))));
```
yravinderkumar33, line 112: 1 - _.sortBy will work fine. But in the case of alphanumeric cases where the initial word is common, it might cause some issue. For example:- Class 1, Class 2, Class 3, Class 4 ......

2- _.capitilize should be changed also in the subscription part. Since the matching is done with casing on. 
A better approach would be to use the angular pipe to capitalize on the initial word.

JagadishPujari: 1. Yeah works for initial word is common 
Screenshot from 2020-04-08 10-32-45

PR #3846, opened 2020-04-09
Issue #SC-1737 fix: Reports download links UI updated

rajeshkumaravel:
- Issue #SC-1737 fix: Reports download links UI updated

---

Requirement:
image

Updated Changes:
image


vinukumar-vs:
Create separate task to refactor duplication of the code in HTML & Angular. Create the map object  & use the values from the object instead of hardcoding the strings in all the places.
@rajeshkumaravel @rajeevsathish

PR #3848, opened 2020-04-09
Issue #SB-18739 filter issue fix

rajeevsathish:
git commit -m Issue


src/app/client/src/app/modules/search/components/course-search/course-search.component.ts
```diff
@@ -69,6 +73,11 @@ export class CourseSearchComponent implements OnInit, OnDestroy, AfterViewInit {
           return of({});
         }
     })).subscribe((filters: any) => {
+      if (this.userService._isCustodianUser && this.orgDetailsFromSlug ) {
```
vinukumar-vs, line 76: Add TODO comment here. So that we can take it later for refactoring this logic.

PR #3854, opened 2020-04-13
Issue #SB-0000 fix:added excryption and decryption for sso login

deveshMantra:


src/app/routes/ssoRoutes.js
```diff
@@ -228,9 +228,13 @@ module.exports = (app) => {
         errType = 'MISSING_QUERY_PARAMS';
         throw 'some of the query params are missing';
       }
-      userName = req.query.id;
+      errType = 'DECRYPTING_DATA';
+      const decryptedData = decrypt(parseJson(decodeURIComponent(req.query.id)));
+      const userData = parseJson(decryptedData);
+      errType = 'VERIFY_TOKEN';
+      verifyToken(userData);
```
vrayulu, line 235: @deveshMantra - why are we trying to reuse the same verifyToken method here also? It will be a problem when you need to change verifyToken method to add/change JWT validations. It will then break this functionality. All we need here is validating expiry timestamp of the decrypted string - it could be a separate small utility function.
@vinukumar-vs

src/app/routes/ssoRoutes.js
```diff
@@ -43,7 +43,7 @@ module.exports = (app) => {
       req.session.userDetails = userDetails;
       logger.info({msg: "userDetails fetched" + userDetails});
       if(!_.isEmpty(userDetails) && (userDetails.phone || userDetails.email)) {
-        redirectUrl = successUrl + getQueryParams({ id: userDetails.userName });
+        redirectUrl = successUrl + encryptUserData({id: userDetails.userName});
```
vrayulu, line 46: @deveshMantra - we added this encryption logic in three places. I am unable to understand the code (due to lack of proper code comments & structuring). I am hoping all the scenarios are addressed. However, we still need refactoring of SSO module to make it modular and easy to maintain.
@vinukumar-vs

src/app/routes/ssoRoutes.js
```diff
@@ -483,6 +487,22 @@ const getQueryParams = (queryObj) => {
...
 }
 
+/**
+ * To generate session for state user logins
+ * using server's time as iat and exp time as 5 min
+ * Session will not be created if exp is expired
+ * @param data object to encrypt data
+ * @returns {string}
+ */
+const encryptUserData = (data) => {
```
vrayulu, line 497: @deveshMantra - we are doing more than just encrypting the data in this method. why are we using JWT format - is it just to reuse verifyToken method or is there any other benefit? 
@vinukumar-vs

PR #3881, opened 2020-04-17
Issue #SB-18833 fix: Course Progress : The content icon color does not turn green when the user clicks on the close icon being in the last page in the PDF content

ArjunKhetia-NIIT:


rajeevsathish:
@ArjunKhetia-NIIT please close this PR this is not the fix for the issue

PR #3883, opened 2020-04-17
Issue #SB-13645 feat: added reCAPTCHA for forgot password workflow

anoophm:


src/app/client/src/app/modules/recover-account/components/identify-account/identify-account.component.ts
```diff
@@ -13,6 +14,8 @@ import { IImpressionEventInput, IEndEventInput, IStartEventInput, IInteractEvent
 export class IdentifyAccountComponent implements OnInit {
 
   disableFormSubmit = true;
+  @ViewChild('captchaRef') captchaRef: RecaptchaComponent;
```
vinukumar-vs, line 17: Write the test-case to make sure ReCaptcha is shown on this page.

PR #3885, opened 2020-04-17
Issue #SB-10681 fix: Added device back button click handler

sudip-mukherjee:


src/app/client/src/app/modules/public/module/player/components/public-content-player/public-content-player.component.ts
```diff
@@ -86,6 +86,13 @@ export class PublicContentPlayerComponent implements OnInit, OnDestroy, AfterVie
    * @memberof ContentPlayerComponent
    */
   ngOnInit() {
+    /** It will handle device back-button click to rotate landscape to portrait */
+    window.addEventListener('orientationchange', () => {
```
anoophm, line 90: @HostListener('window:orientationchange', ['$event']) can be used instead

sudip-mukherjee: @anoophm   Updated the PR

PR #3892, opened 2020-04-18
Issue #SB-16203 fix: District name starting with small letter

JagadishPujari:


src/app/client/src/app/modules/dashboard/components/data-chart/data-chart.component.ts
```diff
@@ -230,6 +230,9 @@ export class DataChartComponent implements OnInit, OnDestroy {
     if (_.get(this.chartConfig, 'labels')) {
       labels = _.get(this.chartConfig, 'labels');
     }
+    _.forEach(labels, (label, key) => {
+      labels[key] = _.capitalize(label);
```
yravinderkumar33, line 234: This looks good, but you need to do the same when showing in filters also.
There also labels should in be title case.

JagadishPujari: That is already fixed and PR is merged

yravinderkumar33: oh okay. Then it's perfect.

src/app/client/src/app/modules/dashboard/components/data-chart/data-chart.component.spec.ts
```diff
@@ -145,7 +145,7 @@ describe('DataChartComponent', () => {
             'Number of QR codes with no linked content': '60'
         }]);
         expect(component.noResultsFound).toBe(false);
-        expect(component.chartLabels).toEqual(['class 2']);
+        expect(component.chartLabels).toEqual(['Class 2']);
```
vinukumar-vs, line 148: This is not a valid test-case.

PR #3905, opened 2020-04-20
Issue #SB-18007 fix: All my contents & collaborations : The contenttype "courseassessment" is not available for the users to apply filters and get the self assess contents

Nik720:
Issue #SB-18007 fix: All my contents & collaborations : The contenttype "courseassessment" is not available for the users to apply filters and get the self assess contents

### DIFF
--- a/src/app/client/src/app/modules/workspace/components/collaborating-on/collaborating-on.component.ts
+++ b/src/app/client/src/app/modules/workspace/components/collaborating-on/collaborating-on.component.ts
@@ -210,17 +210,18 @@ export class CollaboratingOnComponent extends WorkSpace implements OnInit, After
       this.sort = { lastUpdatedOn: this.config.appConfig.WORKSPACE.lastUpdatedOn };
     }
     const preStatus = ['Draft', 'FlagDraft', 'Review', 'Processing', 'Live', 'Unlisted', 'FlagReview'];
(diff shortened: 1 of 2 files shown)


src/app/client/src/app/modules/workspace/components/collaborating-on/collaborating-on.component.ts
```diff
@@ -210,17 +210,18 @@ export class CollaboratingOnComponent extends WorkSpace implements OnInit, After
       this.sort = { lastUpdatedOn: this.config.appConfig.WORKSPACE.lastUpdatedOn };
     }
     const preStatus = ['Draft', 'FlagDraft', 'Review', 'Processing', 'Live', 'Unlisted', 'FlagReview'];
+    // tslint:disable-next-line: max-line-length
+    const contentTypes = bothParams.queryParams.contentType ? bothParams.queryParams.contentType : this.config.appConfig.WORKSPACE.contentType;
```
deveshMantra, line 214: redundant if else can be optimized

Nik720: Resolved

vinukumar-vs:
Have you verified, courseassessment is added in the configuration with can be searchable in the workspace.

PR #3951, opened 2020-04-27
Release 2.10.0 cc

rajeevsathish:


src/app/client/src/app/modules/player-helper/components/content-actions/content-actions.component.html
```diff
@@ -17,10 +17,6 @@
   </div>
 </div>
 
-<app-content-rating #modal *ngIf="contentRatingModal && contentData"
```
vinukumar-vs, line 20: Where is this has implemented in the new design?

PR #3970, opened 2020-04-28
Release 2.10.0 cc

rajeevsathish:


src/app/client/src/app/app.component.ts
```diff
@@ -84,6 +85,8 @@ export class AppComponent implements OnInit, OnDestroy {
   showUserVerificationPopup = false;
   feedCategory = 'OrgMigrationAction';
   labels: {};
+  deviceId: string;
+  uuid = UUID.UUID();
```
vinukumar-vs, line 89: Always correlate the id's. Use portal sessionId as UUID for bot

PR #3985, opened 2020-04-29
Issue #SB-19054 Chat bot application header section is not showing co…

rajeevsathish:
…mpletely


src/app/client/src/app/app.component.ts
```diff
@@ -168,8 +169,9 @@ export class AppComponent implements OnInit, OnDestroy {
     if (this.userService.loggedIn) {
       this.userId = this.userService.userid;
     } else {
-      this.userId = this.userService.anonymousSid;
+      this.userId = this.deviceId;
```
vinukumar-vs, line 172: Why are we using deviceId as UserId. In this case both did & userID will be same. What is the use? What are we not taking anonymous userId here?

PR #3988, opened 2020-04-30
Issue #SB-19181 fix: Location popup in signup page

JagadishPujari:
[JIRA LINK](https://project-sunbird.atlassian.net/browse/SB-19181)


src/app/client/src/app/modules/public/module/signup/signup-routing.module.ts
```diff
@@ -10,6 +10,7 @@ const routes: Routes = [
     path: '', component: SignupComponent,
     data: {
       hideHeaderNFooter : true,
+      hideLocatoionPopup: true,
```
deveshMantra, line 13: please add it in other modules as well recover and signup sign-in

src/app/client/src/app/app.component.ts
```diff
@@ -174,9 +175,15 @@ export class AppComponent implements OnInit, OnDestroy {
...
+  handleLocatoionPopupInSignup() {
+    this.hideLocatoionPopupInSignup = _.get(this.activatedRoute, 'snapshot.firstChild.firstChild.data.hideLocatoionPopup') ||
+          _.get(this.activatedRoute, 'snapshot.firstChild.firstChild.firstChild.data.hideLocatoionPopup');
+  }
+
   isLocationStatusRequired() {
     const url = this.router.url;
-    return !!(_.includes(url, 'signup') || _.includes(url, 'recover') || _.includes(url, 'sign-in'));
+    this.handleLocatoionPopupInSignup();
+    return !!(_.includes(url, 'signup') || _.includes(url, 'recover') || _.includes(url, 'sign-in') || this.hideLocatoionPopupInSignup);
```
deveshMantra, line 186: since it is OR condition hideLocatoionPopupInSignup will always be true for all routes hence never location popup will come for other routes.
Please check it once for other routes

deveshMantra, line 186: not sure if these both ways of checking is required ?

JagadishPujari: No, it will work why because `hideLocatoionPopupInSignup` is undefined in other routes its only true in signup route

src/app/client/src/app/modules/public/module/signup/signup-routing.module.ts
JagadishPujari: Now I have added in signup module and I will add in sign in and recover module

src/app/client/src/app/app.component.ts
JagadishPujari: Since `this.router.url;` is coming undefined if I directly hit on `/signup` so it requires both checks

rajeevsathish:
closing this PR as this is too old and this branch has reached the end of life

PR #4007, opened 2020-05-05
Issue #SB-19188 fix: Added new UI for content player

itsvick:


rajeevsathish:
Coverage is trying to get data for this line of code which is not correct 

templateUrl: './content-player.component.html',
styleUrls: ['./content-player.component.scss']

so merging even though the code coverage is 50%

PR #4022, opened 2020-05-06
Issue #SH-7 feat: As a course creator, I should be able to view the courses created using textbooks under a new content type called curriculumcourse

souravdey091:


harishgilimi:
@souravdey091  This is not covering all the scenarios

PR #4029, opened 2020-05-07
Issue #SH-44 fix : As an unenrolled user, I want to be informed appropriately, when clicking a content on the portal

Nik720:
Issue #SH-44 fix : As an unenrolled user, I want to be informed appropriately, when clicking a content on the portal


src/app/client/src/app/modules/learn/components/course-consumption/course-player/course-player.component.spec.ts
```diff
@@ -478,4 +480,49 @@ describe('CoursePlayerComponent', () => {
     component.contentProgressEvent(telemetryEvent);
     expect(courseConsumptionService.updateContentsState).not.toHaveBeenCalled();
   });
+  it('should show join training popup if course is unenrolled and try to play content', () => {
```
harishgilimi, line 483: Move the duplicate code in both specs to a method and call that method

Nik720: In both specs, there is different code

PR #4045, opened 2020-05-08
Issue #SB-19108 fix: page optimisation

anoophm:


src/app/client/angular.json
```diff
@@ -45,9 +45,12 @@
             "scripts": [
               "src/assets/libs/iziToast/iziToast.min.js",
               "node_modules/jquery/dist/jquery.min.js",
-              "node_modules/slick-carousel/slick/slick.min.js",
               "node_modules/@project-sunbird/telemetry-sdk/index.js",
-              "src/assets/libs/semantic.min.js",
+              "src/assets/libs/semantic-ui/dimmer.min.js",
+              "src/assets/libs/semantic-ui/transition.min.js",
```
vinukumar-vs, line 50: I would like to remove adding more files to this list. This will increase the script file size & portal is loading script file in the initial files list. 
If we do dynamically loading these files depending on the module, then the file size will get reduce. 
example: while do we have to load fancyTree, semantic-ui/modal, for the explore or training page.

PR #4055, opened 2020-05-12
feat #SH-35: As a logged in user, I want to view all my groups on the Portal

Nik720:
- added new module for groups
- added new menu My Groups


src/app/client/src/app/modules/core/components/main-menu/main-menu.component.html
```diff
@@ -81,6 +91,11 @@
 				routerLink="/learn">
 				{{resourceService?.frmelmnts?.tab?.courses}}
 			</a>
+			<a class="item" [class.active]="router.isActive('/search/Groups') || router.isActive('/groups')"
```
harishgilimi, line 94: here also

src/app/client/src/app/modules/core/components/main-menu/main-menu.component.html
```diff
@@ -10,6 +10,10 @@
 			appTelemetryInteract [telemetryInteractEdata]="learnMenuIntractEdata" routerLink="/learn">
 			{{resourceService?.frmelmnts?.tab?.courses}}
 		</a>
+		<a class="item py-8 pl-0 pr-24" [class.active]="router.isActive('/search/Groups') || router.isActive('/groups')"
```
harishgilimi, line 13: what is this **/search/Groups** route ?

src/app/client/src/app/modules/core/components/main-menu/main-menu.component.html
```diff
@@ -110,6 +125,12 @@
 			<a class="item" (click)="showSideBar()" *ngIf="router.isActive('', true)" routerLink="/explore">
 				{{resourceService.frmelmnts?.lbl?.explore}} {{resourceService.instance}}
 			</a>
+			<a class="item"
+				[class.active]="router.isActive('/search/groups') || router.isActive('/groups')"
```
harishgilimi, line 129: what is this **/search/Groups** route ?

src/app/client/src/app/modules/core/components/main-menu/main-menu.component.html
Nik720: if groups can search then it should navigate to this route, I checked other menus and the same way I added.

src/app/client/src/app/modules/core/components/main-menu/main-menu.component.html
harishgilimi: This is not required now

PR #4103, opened 2020-05-18
Issue #SH-67 feat: Adding create user and choose user componenet

souravdey091:


src/app/client/src/app/plugins/profile/profile-routing.module.ts
```diff
@@ -11,6 +11,22 @@ const routes: Routes = [
         env: telemetryEnv, type: 'view', mode: 'create', subtype: 'paginate', object: { type: objectType, ver: '1.0' }
       }, breadcrumbs: [{ label: 'Home', url: '/home' }, { label: 'Profile', url: '' }]
     }
+  },
+  {
+    path: 'create-user', component: CreateUserComponent,
+    data: {
+      telemetry: {
+        env: telemetryEnv, type: 'view', mode: 'create', subtype: 'paginate', object: { type: objectType, ver: '1.0' }
```
harishgilimi, line 19: object is not required here

src/app/client/src/app/plugins/profile/profile-routing.module.ts
```diff
@@ -11,6 +11,22 @@ const routes: Routes = [
...
+      }, breadcrumbs: [{ label: 'Home', url: '/home' }, { label: 'Profile', url: '' }]
+    }
+  },
+  {
+    path: 'choose-user', component: ChooseUserComponent,
+    data: {
+      telemetry: {
+        env: telemetryEnv, type: 'view', mode: 'create', subtype: 'paginate', object: { type: objectType, ver: '1.0' }
```
harishgilimi, line 27: object is not required here

harishgilimi, line 27: mode value is not correct

PR #4110, opened 2020-05-19
Issue #SH-36 feat: As a logged in user, I want to create a group on the Portal

Nik720:
Issue #SH-36 feat: As a logged in user, I want to create a group on the Portal


src/app/client/src/app/app.component.ts
```diff
@@ -525,4 +526,33 @@ export class AppComponent implements OnInit, OnDestroy {
...
+                  host: document.location.origin, // default host
+                  authentication: {
+                      // userToken: string; // optional
+                      // bearerToken: string; // optional
+                  }
+              }
+          },
+          services: {
+              groupServiceConfig: {
+                apiPath: 'api/v1/group',
```
harishgilimi, line 551: apiPath is not correct check other apis which are calling learner service

Nik720: Modified with 'learner/v1/group'

PR #4134, opened 2020-05-21
 Issue #SH-37: added group-actions and member-actions component

swetha-aggidevara:


src/app/client/src/app/modules/groups/services/groups/groups.service.ts
```diff
@@ -85,5 +85,8 @@ export class GroupsService {
     return await this.groupCservice.getAll().toPromise();
   }
 
+  async getGroupById(groupId) {
```
harishgilimi, line 88: Why promise is used here not observables

src/app/client/src/app/modules/groups/components/member-actions/member-actions.component.ts
```diff
@@ -0,0 +1,32 @@
...
+})
+export class MemberActionsComponent implements OnInit {
+  @ViewChild('modal') modal;
+
+  @Input() modalName: string;
+  @Input() member: {};
+  @Output() modalClosed = new EventEmitter();
+  @Output() handleMember = new EventEmitter();
+
+  constructor() {
```
harishgilimi, line 17: Removed the methods which are not required

PR #4169, opened 2020-05-27
Issue #SB-18264 fix:Measure outreach campaigns on the portal

ArjunKhetia-NIIT:


src/app/client/src/app/app.component.ts
```diff
@@ -122,7 +123,42 @@ export class AppComponent implements OnInit, OnDestroy {
...
+            break;
+          case 'utm_content':
+            this.resultJson.push({ 'id': params[item], 'type': 'UtmContent' });
+            break;
+          default:
+            break;
+        }
+      }
+    }
+    sessionStorage.setItem('UTM', JSON.stringify(this.resultJson));
```
anoophm, line 153: If you set this in session when will these be cleared. This param will get added to telemetry for all user action. If user come via explore page second time, then also telemetry will have utm params right. please check this cases.

anoophm, line 153: why are using sessionStorage directly. we have service for this right

src/app/client/src/app/app.component.ts
```diff
@@ -122,7 +123,42 @@ export class AppComponent implements OnInit, OnDestroy {
...
+  }
   ngOnInit() {
+    this.activatedRoute.queryParams.subscribe(params => {
+      const utmParams = ['utm_campaign', 'utm_medium', 'utm_source', 'utm_term', 'utm_content', 'channel'];
```
anoophm, line 157: Am not sure we will get queryParam of the chiled route in rootComponent. Can you verify this

src/app/client/src/app/modules/telemetry/services/telemetry/telemetry.service.ts
```diff
@@ -86,8 +86,9 @@ export class TelemetryService {
    * @param {IStartEventInput} startEventInput
    * @memberof TelemetryService
    */
-  public start(startEventInput: IStartEventInput) {
+  public async start(startEventInput: IStartEventInput) {
```
anoophm, line 89: why is this async function

src/app/client/src/app/modules/telemetry/services/telemetry/telemetry.service.ts
```diff
@@ -86,8 +86,9 @@ export class TelemetryService {
...
     if (this.isInitialized) {
+      startEventInput = await _.cloneDeep(this.addUTM(startEventInput));
```
anoophm, line 91: why do we have await here, does _.cloneDeep returns promise or this.addUTM return promise

src/app/client/src/app/modules/telemetry/services/telemetry/telemetry.service.ts
```diff
@@ -300,4 +309,18 @@ export class TelemetryService {
       raw: window.navigator.userAgent
     };
   }
+
+  public addUTM(object) {
+    if (sessionStorage.getItem('UTM')) {
```
anoophm, line 314: why do we need to get these values from session for every event. Cant we keep these values as a property in telemetry service. create getter and setter for this. Setter can be used to set values for this from app.component also. Why do we need to set value in session in app.component and get value from session in telemetry service.

src/app/client/src/app/app.component.ts
```diff
@@ -122,7 +123,42 @@ export class AppComponent implements OnInit, OnDestroy {
           _.get(this.activatedRoute, 'snapshot.firstChild.firstChild.firstChild.data.hideHeaderNFooter');
       });
   }
+  public makeUTMSession(params) {
+    for (const item in params) {
```
anoophm, line 127: can you write this logic in functional and concise way. Using ENUM to map utm params.
```
enum UTM_PARAMS {
   channel = 'Source';
   utm_campaign = 'Source';
   utm_medium = 'UtmMedium'
}
_.toPairs(param).filter(([key, value]) => UTM_ENUM[key]).map(([key, value]) => ({id: key, value: UTM_ENUM[value]}))
```

src/app/client/src/app/app.component.spec.ts
```diff
@@ -223,4 +223,10 @@ const maockOrgDetails = { result: { response: { content: [{hashTagId: '1235654',
     component.ngOnInit();
     expect(component.showFrameWorkPopUp).toBeTruthy();
   });
+
+  it('Should return proper object by calling make UTM session', () => {
```
anoophm, line 227: this testcase is not correct. This should be triggered from activated route. Note by calling component.makeUTMSession

src/app/client/src/app/app.component.ts
```diff
@@ -86,6 +86,7 @@ export class AppComponent implements OnInit, OnDestroy {
   deviceId: string;
   userId: string;
   appId: string;
+  public resultJson = [];
```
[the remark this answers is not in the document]
ArjunKhetia-NIIT: For testing purpose we need this. Otherwise How can I verify in test case.

src/app/client/src/app/app.component.ts
ArjunKhetia-NIIT: Yes until and unless you explicitly logout or close the tab session won't clear.

ArjunKhetia-NIIT: I had discussed with rajeev he told me why you are using service for this so I am using session storage.

src/app/client/src/app/app.component.ts
ArjunKhetia-NIIT: Yes verified it's not only for root if we pass the parameter in any route from any place I am capturing that and processing as per requirement.

src/app/client/src/app/modules/telemetry/services/telemetry/telemetry.service.ts
ArjunKhetia-NIIT: because I am making modification in each and every data formation call so waiting for the data to format well.

src/app/client/src/app/modules/telemetry/services/telemetry/telemetry.service.ts
ArjunKhetia-NIIT: Yes I am waiting for data to reform it properly.

src/app/client/src/app/app.component.ts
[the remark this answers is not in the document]
anoophm: cant you just put a spy on sessionStorage setItems and check whether its called with param. Please think of other options also.

src/app/client/src/app/app.component.ts
anoophm: ok, please check all cases once. Before going with session

rajeevsathish: hey If I had said that I am really sorry we should use the service for session store please don't use directly

src/app/client/src/app/app.component.ts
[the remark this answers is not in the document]
ArjunKhetia-NIIT: Done made it const instead of public variable

src/app/client/src/app/app.component.ts
ArjunKhetia-NIIT: Done with storing inside telemetry service and making 1 copy to session storage also.

src/app/client/src/app/modules/telemetry/services/telemetry/telemetry.service.ts
ArjunKhetia-NIIT: Done with storing inside telemetry service and making 1 copy to session storage also.

src/app/client/src/app/app.component.ts
ArjunKhetia-NIIT: Done implementing as you suggested.

src/app/client/src/app/app.component.spec.ts
ArjunKhetia-NIIT: Done modification for the same in test cases.

PR #4170, opened 2020-05-27
Issue #SB-19351 fix: create mentor can create only one batch on multi…

swetha-aggidevara:
…ple clicks of create


swetha-aggidevara:
> this.disableSubmitBtn = true; use this variable to disable the button so that click event will not be triggered

this.diableSubmitBtn is not disabling the button

PR #4199, opened 2020-05-29
Issue 18983 fix: lazzy loading scripts when required

anoophm:


.circleci/config.yml
```diff
@@ -18,18 +18,13 @@ jobs:
...
           name: Installing angular cli
-          command: 'npm install -g @angular/cli@6.0.8'
-      - restore_cache:
-          key: dependency-cache-{{ checksum "package.json" }}
+          command: 'npm install -g @angular/cli@6.0.8 --no-progress'
       - run:
           name: Installing npm deps
-          command: 'npm install'
-      - save_cache: # special step to save the dependency cache
-          key: dependency-cache-{{ checksum "package.json" }}
```
rjshrjndrn, line 28: Why are we removing the caching process ? @anoophm

anoophm: will revert this back. @rajeevsathish suggested this was done last time we they faced the issue. But this dint solve the issue

PR #4228, opened 2020-06-02
Issue #SB-19079 fix: Code sanitize for removal of editors from portal

rajeshkumaravel:
- Issue #SB-19079 fix: Code sanitize for removal of editors from portal


src/app/routes/clientRoutes.js
```diff
@@ -165,9 +165,9 @@ function getLocals(req) {
   locals.deviceId = '';
   locals.deviceProfileApi = envHelper.DEVICE_PROFILE_API;
   locals.slug = slug ? slug : '';
-  locals.collectionEditorCDN = envHelper.CONTENT_EDITORS_CDN_URL.COLLECTION_EDITOR;
-  locals.contentEditorCDN = envHelper.CONTENT_EDITORS_CDN_URL.CONTENT_EDITOR;
-  locals.genericEditorCDN = envHelper.CONTENT_EDITORS_CDN_URL.GENERIC_EDITOR;
+  locals.collectionEditorURL = envHelper.CONTENT_EDITORS_URL.COLLECTION_EDITOR;
```
anoophm, line 168: we can keep one base path and change relative path based on type of editor. These reduces number of env variable we need to maintain. 
Base: https://domain/editor/ :- this will be our base path and env variable in portal backend

relative paths for 3 editors
1. collection: collection-editor/index.html
2. content: content-editor/index.html
3. generic: generic-editor/index.html
these three can be kept in editor.config.json

@rajeevsathish @rajeshkumaravel let me know if this is possible.

rajeshkumaravel: We do not have relative path for editors.
We are referring BLOB URL.

PR #4283, opened 2020-06-08
Issue #SB-18846 fix: Added userData to playerConfig for only logged in user

rajeshkumaravel:


src/app/client/src/app/modules/player-helper/components/player/player.component.ts
```diff
@@ -82,6 +84,23 @@ export class PlayerComponent implements OnInit, AfterViewInit, OnChanges {
...
       }
     }
+    // Check for loggedIn user; and append user data to context object
+    // User data (`firstName` and `lastName`) is used to show at the end of quiz
+    if (this.playerConfig && this.userService.loggedIn) {
+      this.userService.userData$.subscribe((user: any) => {
+        if (user && !user.err) {
+          const userProfile = user.userProfile;
+          this.playerConfig.context['userData'] = {
+            identifier: userProfile.identifier ? userProfile.identifier : '',
```
rajeevsathish, line 94: Why do we need userName to be sent
and also what is the use of an identifier?

PR #4308, opened 2020-06-11
Issue #SB-19488 fix: Added content routes for whitelisting

rajeshkumaravel:
- Issue #SB-19488 fix: Added content routes for whitelisting


src/app/helpers/apiWhiteList.js
```diff
@@ -0,0 +1,52 @@
...
+ * 2. If API is does not exists then request is terminated and 401 UNAUTHORIZED response is sent
+ * @since release-3.1.0
+ */
+const isAPIWhiteListed = () => {
+  return function (req, res, next) {
+    const REQ_URL = req.originalUrl;
+    if (_.get(ROUTE, REQ_URL)) {
+      next();
+    } else {
+      logger.error({
```
vinukumar-vs, line 26: Use UtilService.logError method for standard portal log format

src/app/helpers/apiWhiteList.js
```diff
@@ -0,0 +1,52 @@
...
+const dateFormat  = require('dateformat');
+const ROUTE       = require('./whitelistApis');
+
+/**
+ * @description - Function to check whether incoming API is whitelisted (or) not
+ * 1. If incoming API is present in API_LISTS (whitelistApis); then request is passed further
+ * 2. If API is does not exists then request is terminated and 401 UNAUTHORIZED response is sent
+ * @since release-3.1.0
+ */
+const isAPIWhiteListed = () => {
```
vinukumar-vs, line 20: Change the name to isValid or IsExist

src/app/routes/contentRoutes.js
```diff
@@ -37,6 +38,7 @@ module.exports = (app) => {
         }))
 
     app.all('/content/*',
+        isAPIWhitelisted.isAPIWhiteListed(),
```
vinukumar-vs, line 41: after naming changes, it should be like
apiWhiteList.isValid()

src/app/helpers/apiWhiteList.js
rajeshkumaravel: I would recommend having method name as **isAllowed**; which conveys whether API is allowed or not.
Please confirm on this.
_P.S. Awaiting response._

src/app/helpers/apiWhiteList.js
rajeshkumaravel: As a reference from whitelistUrl.js the logger used is ```const logger = require('sb_logger_util_v2');```
And also the UtilService.logError is nowhere used;

vinukumar-vs: Use logError method going foward.

rajeshkumaravel: Changes updated.

src/app/helpers/apiWhiteList.js
rajeshkumaravel: Changes updated.

PR #4309, opened 2020-06-11
Issue #SH-312 feat: Review T&C re-acceptance workflow for LIU and MAU

sudip-mukherjee:


src/app/client/src/app/modules/shared-feature/components/tnc-popup/terms-conditions-popup.component.ts
```diff
@@ -75,7 +75,7 @@ export class TermsAndConditionsPopupComponent implements OnInit, OnDestroy {
       }
     };
     if (this.userService.userProfile.managedBy) {
-      requestBody.request['userId'] = this.userService.userProfile.managedBy;
+      requestBody.request['userId'] = this.userService.userid;
```
sudip-mukherjee, line 78: @deveshMantra please review this segment.

deveshMantra: Its okay as in release-3.0.0 we are using same code seems like 3.0.0 is not merged with 3.1.0.
Sending managed-user id for TNC accept and not sneding user id if it is parent user.

src/app/client/src/app/app.component.ts
```diff
@@ -290,12 +290,14 @@ export class AppComponent implements OnInit, OnDestroy {
...
-      this.showTermsAndCondPopUp = true;
-    } else {
-      this.checkFrameworkSelected();
-    }
+    this.userService.userData$.subscribe((user: IUserData) => {
+      if (_.has(this.userService.userProfile, 'promptTnC') && _.has(this.userService.userProfile, 'tncLatestVersion') &&
+        _.has(this.userService.userProfile, 'tncLatestVersion') && this.userService.userProfile.promptTnC === true) {
+        this.showTermsAndCondPopUp = true;
+      } else {
+        this.checkFrameworkSelected();
```
deveshMantra, line 298: just cross check if it will not call device location API multiple time on every switch

src/app/client/src/app/modules/shared-feature/components/tnc-popup/terms-conditions-popup.component.ts
```diff
@@ -75,7 +75,7 @@ export class TermsAndConditionsPopupComponent implements OnInit, OnDestroy {
       }
     };
     if (this.userService.userProfile.managedBy) {
```
harishgilimi, line 77: this check is not required

deveshMantra: we have confirmed with backed team we should only send the user id for child users only.
If we need to send user id always(for child and managed users) we might need to change SSO sign-up and self sign up flows as well to send user id always to have consistency.

PR #4330, opened 2020-06-12
Issue #SB-19202 Add Chat bot to portal service

rajeevsathish:


rajeevsathish:
Merging this PR as only the variables are added so we cannot write test case for the same

PR #4332, opened 2020-06-12
Issue #SH-393 fix:user read api should be called with token

deveshMantra:


src/app/proxy/proxyUtils.js
```diff
@@ -35,6 +35,10 @@ const decorateRequestHeaders = function () {
     if(!srcReq.get('X-App-Id')){
       proxyReqOpts.headers['X-App-Id'] = appId
     }
+    if (srcReq.session.managedToken) {
+      proxyReqOpts.headers['x-authenticated-for'] = srcReq.session.managedToken
```
deveshMantra, line 39: this key is correct?

PR #4368, opened 2020-06-18
Issue #SB-0000 fix:changed the headers forwarding

deveshMantra:


rajeevsathish:
#SB-19616 change the device profile api to send the X-REAL-IP

PR #4390, opened 2020-06-22
Issue #SB-19033 feat: Add publisher in board filter

souravdey091:


rajeevsathish:
TODO @souravdey091  you need to remove the hardcoding of the publisher and also the component to be moved to SB-forms repo

PR #4504, opened 2020-07-03
Issue #SB-19724 fix:location popup should not come

deveshMantra:


src/app/client/src/app/app.component.ts
```diff
@@ -234,7 +234,15 @@ export class AppComponent implements OnInit, OnDestroy {
           }
         }
       }
-      this.showUserTypePopup = !localStorage.getItem('userType');
+      // TODO: code can be removed in 3.1 release from user-onboarding component as it is handled here.
+      zip(this.tenantService.tenantData$, this.orgDetailsService.orgDetails$).subscribe((res) => {
```
vinukumar-vs, line 238: Revisit this logic in the next release properly. I can see multiple places subscribe of orgDetailsService & tenantService. We have to keep the workflow proper at one place for loggedIn or non-loggedIn user.

rajeevsathish, line 238: getOrgDetails, TenantService subscribed  multiple places as per @vinukumar-vs please check @deveshMantra

PR #4528, opened 2020-07-07
Issue #TG-259 fix: google SignIn state information stored in session instead of state query param that gets sent to google

anoophm:
…instead of state query param that gets sent to google

### DIFF
--- a/src/app/routes/googleSignInRoutes.js
+++ b/src/app/routes/googleSignInRoutes.js
@@ -6,6 +6,12 @@ const logger = require('sb_logger_util_v2')
 const utils = require('../helpers/utilityService');
 const GOOGLE_SIGN_IN_DELAY = 3000;
 const uuid = require('uuid/v1')
+/**
+ * keycloack adds this string to track auth redirection and 
+ * with this it triggers auth code verification to get token and create session
+ * google flow this is not required
+ */
+const KEYCLOACK_AUTH_CALLBACK_STRING = 'auth_callback=1'; 
 
 module.exports = (app) => {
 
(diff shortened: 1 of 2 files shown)


src/app/routes/googleSignInRoutes.js
```diff
@@ -161,3 +172,33 @@ const handleGoogleProfileError = (error) => {
   });
   throw error.error || error.message || error;
 };
+
+var handleGoogleAuthEvent = () => {
```
vinukumar-vs, line 176: Where are we using this function?

src/app/routes/googleSignInRoutes.js
```diff
@@ -58,9 +69,9 @@ module.exports = (app) => {
       keyCloakToken = await createSession(googleProfile.emailId, reqQuery, req, res).catch(handleCreateSessionError);
       logger.info({msg: 'keyCloakToken fetched' + JSON.stringify(keyCloakToken)});
       errType = 'UNHANDLED_ERROR';
-      redirectUrl = reqQuery.redirect_uri.split('?')[0];
+      redirectUrl = reqQuery.redirect_uri.replace(KEYCLOACK_AUTH_CALLBACK_STRING, ''); // to avoid 401 auth errors from keycloak
       if (reqQuery.client_id === 'android') {
-        redirectUrl = redirectUrl + getQueryParams(keyCloakToken);
+        redirectUrl = reqQuery.redirect_uri.split('?')[0] + getQueryParams(keyCloakToken);
```
vinukumar-vs, line 74: This code will fail in mobile use-case right. When there are UTM params in URL. This has to verify on mobile.

PR #4530, opened 2020-07-07
Issue #SB-19675 fix: Added captcha for generate OTP at forgot password, select account and resend otp at verify account page.

rajeshkumaravel:
- Issue #SB-19675 fix: Added captcha for generate OTP at forgot password, select account and resend otp at verify account page.


rajeshkumaravel:
- Overall coverage
image

image

image

image

src/app/client/src/app/modules/public/module/signup/services/signup/signup.service.ts
```diff
@@ -20,6 +20,14 @@ export class SignupService {
     return this.learnerService.post(options);
   }
 
+  generateOTPforAnonymousUser(data, captchaResponse) {
+    const options = {
+      url: this.configService.urlConFig.URLS.OTP.ANONYMOUS.GENERATE + '?captchaResponse=' + captchaResponse,
```
vinukumar-vs, line 25: Why are we using new route for anonymous. Generate OTP should be common for any use-case.

PR #4593, opened 2020-07-14
Issue #SH-858 fix: added new route in whitelistApi

Nik720:


src/app/helpers/whitelistApis.js
```diff
@@ -205,6 +205,10 @@ const API_LIST = {
     '/learner/data/v1/system/settings/get/:slug': {
       checksNeeded: []
     },
+    '/learner/data/v1/system/settings/get/googleReCaptcha': {
+      checksNeeded: ['ROLE_CHECK'],
```
rajeevsathish, line 209: is the person who will use this will be loggedin if not then the ROLE_CHECK will not work

Nik720: Yes, the user is logged in user

rajeevsathish:
Approved

PR #4606, opened 2020-07-15
Issue #SH-681 fix: Batch Details is displayed in the Pop up Post enrolment End date

souravdey091:


src/app/client/src/app/modules/learn/components/batch/batch-details/batch-details.component.html
```diff
@@ -10,17 +10,17 @@
                 <div>{{resourceService?.frmelmnts?.lbl?.availableBatches}}</div>
                 <div>{{resourceService?.frmelmnts?.lbl?.courseCreatedBy}}</div>
             </div>
-            <div class="join-batch__body" *ngFor="let batch of allBatchList">
+            <span  *ngFor="let batch of allBatchList" >
+                <div class="join-batch__body" *ngIf="!isEnrollmentAllowed(batch?.enrollmentEndDate)">
```
harishgilimi, line 14: Remove this in ts file itself

souravdey091: Can't remove it as batch creator needs to see the batch whose enrollment date is over

harishgilimi: you should differentiate assign it to another property with map function and use that here

souravdey091: Done

PR #4641, opened 2020-07-20
Issue #SH-252 fix: Changing en.properties, fixing back url, disabling join course when no batch available

souravdey091:


src/app/client/src/app/modules/learn/services/course-consumption/course-consumption.service.ts
```diff
@@ -19,6 +19,7 @@ export class CourseConsumptionService {
   launchPlayer = new EventEmitter<any>();
   updateContentState = new EventEmitter<any>();
   showJoinCourseModal = new EventEmitter<any>();
+  disableJoinCourseBtn = new EventEmitter();
```
harishgilimi, line 22: instead of  **disableJoinCourseBtn** it should be **enableCourseEntrollment** and emit for both  when openBatchCount is 0 as **false** if more than one then as **true**. use this directly in html template with help of async pipe

PR #4687, opened 2020-07-24
Issue #SB-19207 fix: Wrong logo gets displayed when redirected from one tenant to the root tenant in explore page

VenkatKadiveti:


src/app/client/src/app/app.component.ts
```diff
@@ -416,11 +416,11 @@ export class AppComponent implements OnInit, OnDestroy {
         this.orgDetails = data;
         this.channel = this.orgDetails.hashTagId;
         this.botObject['channel'] = this.channel;
-        if (this.userService.slug !== '') {
+        // if (this.userService.slug !== '') {
```
rajeevsathish, line 419: what happen if the slug value is not there you are still saving the slug with empty string?

PR #4695, opened 2020-07-28
Issue #SB-19927 fix: Added encryption and decryption for sso flow.

rajeshkumaravel:


src/app/routes/accountRecoveryRoute.js
```diff
@@ -73,14 +73,18 @@ module.exports = (app) => {
...
             }
             if (req.body.request.userId) {
-              encrypt['id'] = req.body.request.userId
+              _encrypt['id'] = req.body.request.userId
             }
             var timeInMin = 5;
-            var validator = encriptWithTime(encrypt, timeInMin);
+            var validator = encriptWithTime(_encrypt, timeInMin);
+            const reqType = req.body.request.type;
+            const dataToEncrypt = {};
```
deveshMantra, line 85: shall we be using expirating time as we did for other places ?

src/app/routes/accountRecoveryRoute.js
```diff
@@ -73,14 +73,18 @@ module.exports = (app) => {
...
+            dataToEncrypt[reqType] = req.body.request.key;
+            req.session.otpEncryptedInfo = encrypt(JSON.stringify(dataToEncrypt));
```
deveshMantra, line 87: is two times encryption is required ?
can we include if condition like for user id ?

src/app/routes/accountRecoveryRoute.js
rajeshkumaravel: No.
We are not using time expiry.

src/app/routes/accountRecoveryRoute.js
rajeshkumaravel: `if` condition on what ?
type is always available in verify OTP route.

PR #4708, opened 2020-07-29
Issue #SB-19808 fix:added new UI for resources and explore components

deveshMantra:


src/app/client/src/app/modules/core/components/main-menu/main-menu.component.html
```diff
@@ -146,17 +146,10 @@
...
-			appTelemetryInteract [telemetryInteractEdata]="libraryMenuIntractEdata" routerLink="/resources">
-			<span class="sb-icon sb-icon__textbook sb-icon--md mr-4"></span>
-			{{resourceService?.frmelmnts?.tab?.resources}}
-		</a>
-		<a class="item item--course" [class.active]="router.isActive('/search/Courses') || router.isActive('/learn')"
-			appTelemetryInteract [telemetryInteractEdata]="learnMenuIntractEdata" routerLink="/learn">
-			<span class="sb-icon sb-icon__course sb-icon--md mr-4"></span>
-			{{resourceService?.frmelmnts?.tab?.courses}}
-		</a>
+  <ng-container *ngIf="userService.loggedIn">
```
rajeevsathish, line 149: I think the logged and non logged in user will have same menu so we do not need this logic here

PR #4758, opened 2020-08-05
Issue #SB-18683 fix: Resend otp and submit otp button text is not aligned properly in OTP page when user selects  tamil language

VenkatKadiveti:


VenkatKadiveti:
created new PR

PR #4910, opened 2020-08-14
 Issue #SH-638 fix: added backend route for issuing certificate

swetha-aggidevara:


src/app/helpers/certHelper.js
```diff
@@ -90,4 +91,36 @@ const getUserEnrolledCourses = async (req, courseId, userId) => {
...
+      next(err);
+    }
+}
+}
+
+const getTemplateData = async (req, requestParams) => {
+  logger.info({msg: `getTemplateData() is called`});
+  const appConfig = getHeaders();
+  appConfig.headers['x-authenticated-user-token'] = _.get(req, 'kauth.grant.access_token.token') || _.get(req, 'headers.x-authenticated-user-token');
+  const response = await HTTPService.post(`${certRegURL + 'org/v2/preferences/read'}`, {request: requestParams}, appConfig).toPromise();
```
vinukumar-vs, line 121: User common service method instead of adding the HTTPService.post

Example:
const { sendRequest } = require('./httpRequestHandler');
googleservice.js -> line 35

src/app/helpers/certHelper.js
```diff
@@ -90,4 +91,36 @@ const getUserEnrolledCourses = async (req, courseId, userId) => {
...
+        next();
+    } catch(err) {
+      logError(req, err, 'Error occurred while fetching template');
+      next(err);
+    }
+}
+}
+
+const getTemplateData = async (req, requestParams) => {
+  logger.info({msg: `getTemplateData() is called`});
```
vinukumar-vs, line 118: Change logger.info messages to somthing like below(plain english statement) 
Ex: logger.info("Adding certificate {} to a batch id {}", req.get("templateid"), req.get(batchid));

PR #4933, opened 2020-08-16
Issue #SH-640 fix: Added toast message and template change detection modal

sudip-mukherjee:


gandham-santhosh:
Please have the unit test cases added before it goes to preprod

PR #5008, opened 2020-08-20
Issue #SB-19877 fix: Added test case for Navigation service

rajeshkumaravel:


src/app/client/src/app/modules/shared/services/navigate/router-navigation.service.spec.ts
```diff
@@ -1,17 +1,50 @@
...
+        url: [
+          {
+            path: '/home'
+          }
+        ]
+      }
+    };
+    spyOn(service, 'navigateToParentUrl').and.returnValue(of({}));
+    service.navigateToParentUrl(routeMock);
+    expect(service.navigateToParentUrl).toHaveBeenCalled();
```
vinukumar-vs, line 47: Actual assertion should be you have to check for "component.parentUrl" having the right value what you are expecting.

vinukumar-vs (changes requested):

src/app/client/src/app/modules/shared/services/navigate/router-navigation.service.spec.ts
rajeshkumaravel: Changes have been updated.

vinukumar-vs (approved):

vinukumar-vs (approved):

PR #5088, opened 2020-09-02
Issue #SH-996 feat: Portal: Do not allow user to enroll to future batches but prevent them from consumption

VenkatKadiveti:


src/app/client/src/app/modules/learn/components/batch/batch-details/batch-details.component.ts
```diff
@@ -175,7 +175,14 @@ export class BatchDetailsComponent implements OnInit, OnDestroy {
           });
           // If batch length is 1, then directly join the batch
           if (this.allBatchList && this.allBatchList.length === 1) {
-            this.enrollBatch(this.allBatchList[0]);
+            const batchStartDate = new Date(this.allBatchList[0]['startDate']);
```
rajeevsathish, line 178: why do we need the same chk in this and also int he enroll batch call?

src/app/client/src/app/modules/learn/components/course-consumption/course-player/course-player.component.ts
```diff
@@ -434,9 +437,45 @@ export class CoursePlayerComponent implements OnInit, OnDestroy {
       this.router.navigate(['/learn/course/play', collectionUnit.identifier], navigationExtras);
     } else {
       this.showJoinTrainingModal = true;
+      const currentDate = new Date();
+      if(this.courseHierarchy.batches && this.courseHierarchy.batches.length === 1){
```
rajeevsathish, line 441: I feel this is not the optimized code and also if startDate is not there then the js will fail

rajeevsathish (changes requested):

src/app/client/src/app/modules/learn/components/batch/batch-details/batch-details.component.ts
VenkatKadiveti: If the course has only one batch then we are not opening the Pop up and we are performing direct enrollment to the course for that ongoing batch but if That one batch is future batch then we are checking here.

VenkatKadiveti: Got It , I update the PR

PR #5100, opened 2020-09-04
Issue #sh-918 feat: Collection dashboard - summary aggregates

VenkatKadiveti:


src/app/client/src/app/modules/shared/components/sb-datatable/sb-datatable.component.ts
```diff
@@ -0,0 +1,85 @@
...
+    } else {
+      this.tableData = _.filter(this.data, (row) => {
+        return (_.lowerCase(JSON.stringify(row))).includes(_.lowerCase(this.searchData))
+      })
+    }
+
+  }
+  filterData(data) {
+    let resultData = [];
+    _.forEach(data, (row, index) => {
```
vinukumar-vs, line 49: User array or library filter method or create the angular pipe to filter.
Dont use, nested forEach loop. This will cause performance issue.

vinukumar-vs (changes requested):

PR #5111, opened 2020-09-07
Issue #SB-18924 fix: removed sb_logger_util_v2

Nik720:


src/app/helpers/apiWhiteList.js
```diff
@@ -69,12 +69,12 @@ const urlChecks = {
    * @since - release-3.1.0
    */
   ROLE_CHECK: (resolve, reject, req, rolesForURL, REQ_URL) => {
-    logger.info({
```
Nik720, line 72: **logger.info** giving error and because of this we are not able to login, so comment this

rajeevsathish (approved):

PR #5146, opened 2020-09-14
Issue SB-20618 fix : resource of contentType lessonplan linked to any collection on deletion should display error dialog

rajnish-dargan:


rahulshukla (approved):
changed from content type to mime type to figure out ETB and course

pallakartheekreddy (approved):

PR #5148, opened 2020-09-15
Issue #SB-20620 fix: Group service issue to play the content

vinukumar-vs:


src/app/client/src/app/modules/shared/directives/add-to-group/add-to-group.directive.ts
```diff
@@ -14,7 +14,10 @@ import { TelemetryService } from '@sunbird/telemetry';
 
 // tslint:disable-next-line:only-arrow-functions
 export function csGroupServiceFactory() {
-  return CsModule.instance.groupService;
+  if(CsModule.instance.isInitialised){
```
rajeevsathish, line 17: this may give error if the instance object is not there so JS error may come

rajeevsathish: can you please use _.get so that undefined is handled

rajeevsathish (changes requested):

rajeevsathish (approved):

PR #5157, opened 2020-09-17
Issue #SH-771 feat: generalisation of labels in UI for end users

Nik720:


src/app/server.js
```diff
@@ -203,6 +204,8 @@ require('./proxy/localProxy.js')(app) // Local proxy for content and learner ser
 
 app.all('/v1/user/session/create', (req, res) => trampolineServiceHelper.handleRequest(req, res, keycloak))
 
+app.get('/getGeneralisedResourcesBundles/:lang/:fileName', proxyUtils.addCorsHeaders, getGeneralisedResourcesBundles);
```
rajeevsathish, line 207: why do we need lang here?

rajeevsathish (approved):

PR #5181, opened 2020-09-19
Issue #SB-20667 fix:updated interface and added query param for conte…

Dhirenaade:
…nt read.


rahulshukla (approved):
Adding Primary and Additional categories in read query params

pallakartheekreddy (approved):

PR #5191, opened 2020-09-21
Issue #sh-1115 fix : As a state admin, I want to agree to course t&c while creating a batch

VenkatKadiveti:


src/app/client/src/app/modules/shared/services/config/url.config.json
```diff
@@ -184,7 +184,8 @@
       "GET_DETAILS": "course/v1/batch/read",
       "GET_PARTICIPANT_LIST": "course/v1/batch/participants/list",
       "GET_BATCHS": "course/v1/batch/list",
-      "REMOVE_USERS": "course/v1/batch/user/remove"
+      "REMOVE_USERS": "course/v1/batch/user/remove",
+      "TERMS_AND_CONDITIONS": "https://docs.google.com/document/d/1QLtCVlHR_AELMmLDONxSekE__M-4sZaSe8pzpBj3g0U/edit"
```
vinukumar-vs, line 188: This is not the actual URL. This will vary for every environment. Do it similar to groups

vinukumar-vs (changes requested):

vinukumar-vs (approved):

PR #5251, opened 2020-09-29
Issue #sh-918 fix: loading ondemad reports while changing the batch

VenkatKadiveti:


src/app/client/src/app/modules/shared/components/on-demand-reports/on-demand-reports.component.ts
```diff
@@ -58,6 +58,9 @@ export class OnDemandReportsComponent implements OnInit {
       });
     }
   }
+  ngOnChanges(){
+    this.loadReports();
```
vinukumar-vs, line 62: Why are loading always.. It should load only when the users is in DetailsReport tab

vinukumar-vs (changes requested):

vinukumar-vs (approved):

PR #5278, opened 2020-10-05
Issue #HE-50 feat: map component added as new chart Type in Portal

yravinderkumar33:


src/app/client/src/app/modules/dashboard/components/map/map.component.ts
```diff
@@ -0,0 +1,311 @@
...
+    lonBounds: [35.513327, 97.39535869999999],
+    initialZoomLevel: 5,
+    controlTitle: 'India Heat Map',
+    tileLayer: {
+      urlTemplate: 'https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png',
+      options: {
+        attributions: '&copy; <a href="https://www.openstreetmap.org/copyright">OpenStreetMap</a> contributors'
+      }
+    },
+    rootStyle: {
```
rajeevsathish, line 39: why is the css in the ts file?

rajeevsathish (changes requested):

src/app/client/src/app/modules/dashboard/components/map/map.component.ts
yravinderkumar33: No, it's not CSS. These are attributions that are required by the leaflet to render its style. By default, I've set open streets but it's configurable and can be changed via the report configuration.
and rest others are default styling configurations passed for the leaflet control box which is also configurable.
If we don't pass the configuration then it'll take these default configurations.

rajeevsathish (approved):

PR #5384, opened 2020-10-28
Issue #SH-1300 fix: copy  etb as course

swetha-aggidevara:


src/app/client/src/app/modules/shared-feature/components/collection-player/collection-player.component.ts
```diff
@@ -601,6 +605,35 @@ export class CollectionPlayerComponent implements OnInit, OnDestroy, AfterViewIn
     });
   }
 
+  getChildren(collection) {
```
vinukumar-vs, line 608: This will end up in infinite loop if there is any issue in code. We should not write calling the same function by itself based on condition.

vinukumar-vs (changes requested):

src/app/client/src/app/modules/shared-feature/components/collection-player/collection-player.component.ts
swetha-aggidevara: Without recursive function i won't get children data to filter those fields

swetha-aggidevara: and it won't end up infinite loop, since we are returning  data

src/app/client/src/app/modules/shared-feature/components/collection-player/collection-player.component.ts
```diff
@@ -601,6 +604,23 @@ export class CollectionPlayerComponent implements OnInit, OnDestroy, AfterViewIn
     });
   }
 
+  getOnlyRequiredProperties(collection) {
```
vinukumar-vs, line 607: Take 'requiredKeys' also as an input to this method. So that anybody can use this method to parse the collection.

Write this method in utils & call that from 
`
`
const requiredProps = ['mimeType', 'visibility', 'identifier', 'selected', 'name', 'contentType', 'children',
    'primaryCategory', 'additionalCategory', 'parent'];
collection = utils.reduceTreeProps(collection, requiredProps);
`

vinukumar-vs (approved):

vinukumar-vs (approved):

PR #5442, opened 2020-11-09
Issue #SB-21312 fix: download members progress as csv

swetha-aggidevara:


src/app/client/src/app/modules/groups/components/activity/activity-dashboard/activity-dashboard.component.ts
```diff
@@ -247,6 +256,34 @@ export class ActivityDashboardComponent implements OnInit, OnDestroy {
     return _.lowerCase(_.get(this.queryParams, 'title'));
   }
 
+  downloadCSVFile() {
```
vinukumar-vs, line 259: Why this is again part of component(SRP). This should be utils functions, You pass array/collection data & what fields to be added into csv file.
downloadCSV(Collection, fiekds)

vinukumar-vs (changes requested):

vinukumar-vs (approved):

PR #5471, opened 2020-11-18
Issue sh-1399 fix: On clicking the Edit Certificate, the attached cer…

VenkatKadiveti:
…tificate template is not displayed

### DIFF
--- a/src/app/client/src/app/modules/certificate/components/certificate-configuration/certificate-configuration.component.html
+++ b/src/app/client/src/app/modules/certificate/components/certificate-configuration/certificate-configuration.component.html
@@ -185,7 +185,7 @@
               <div class="certificate-edit-sbcard">
                 <div class="d-flex flex-dc p-16 certificate-edit-sbcard-content">
                   <div class="certificate-frame">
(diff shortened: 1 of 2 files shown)


vinukumar-vs (approved):

src/app/client/src/app/modules/certificate/components/certificate-configuration/certificate-configuration.component.html
```diff
@@ -185,7 +185,7 @@ <h3>{{resourceService?.frmelmnts?.cert?.lbl?.noEditAction}}</h3>
               <div class="certificate-edit-sbcard">
                 <div class="d-flex flex-dc p-16 certificate-edit-sbcard-content">
                   <div class="certificate-frame">
-                    <div><img [src]="selectedTemplate?.artifactUrl"></div>
+                    <div><img src="{{selectedTemplate?.previewUrl || selectedTemplate?.artifactUrl}}"></div>
```
vinukumar-vs, line 188: Do reverse, if artifactUrl is not there then user previewUrl. 
Ideally we should not use previewUrl at all. This is only for release-3.3 we used.

vinukumar-vs (approved):

PR #5481, opened 2020-11-19
SH-1229:  certificate

swetha-aggidevara:


src/app/client/src/app/modules/learn/services/course-consumption/course-consumption.service.ts
```diff
@@ -185,6 +186,11 @@ getAllOpenBatches(contents) {
           }
       });
       const visibility: boolean = mentorBatches ? mentorBatches.length > 0 : false;
+      this._userCreatedAnyBatch = visibility;
       this.userCreatedAnyBatch.emit(visibility);
   }
+
+  get isUserCreatedAnyBatch() {
```
vinukumar-vs: Name is not indicating the right action

src/app/client/src/app/modules/learn/services/course-consumption/course-consumption.service.ts
```diff
@@ -23,7 +23,8 @@ export class CourseConsumptionService {
   showJoinCourseModal = new EventEmitter<any>();
   enableCourseEntrollment = new EventEmitter();
   coursePagePreviousUrl: any;
-  userCreatedAnyBatch = new EventEmitter();
+  public _isMentorOfAnyBatch = false;
```
vinukumar-vs, line 26: You should not add this variable here. It should be similar to canCreateBatch() method.

src/app/client/src/app/modules/learn/services/course-consumption/course-consumption.service.ts
```diff
@@ -23,7 +23,8 @@ export class CourseConsumptionService {
...
+  public userCreatedAnyBatch = new EventEmitter();
```
vinukumar-vs, line 27: This emitter is already giving the data. Why are we using new method now?

vinukumar-vs (changes requested):

vinukumar-vs (approved):

PR #5482, opened 2020-11-19
Issue #SB-20299 feat: Integrating video player with portal

souravdey091:


src/app/client/src/app/modules/player-helper/components/player/player.component.html
```diff
@@ -29,12 +29,14 @@
...
-        <sunbird-pdf-player 
-        [playerConfig]="playerConfig" (playerEvent)="pdfEventHandler($event)"
-        (telemetryEvent)="generateContentReadEvent($event, true)"
-        ></sunbird-pdf-player>
+      <div class="contentViewerIframeShadow" [ngClass]="{'player-fullscreen content-full-video': isFullScreenView}">
+        <sunbird-pdf-player *ngIf="playerConfig.metadata.mimeType === 'application/pdf'" [playerConfig]="playerConfig"
+          (playerEvent)="eventHandler($event)" (telemetryEvent)="generateContentReadEvent($event, true)">
+        </sunbird-pdf-player>
+
+        <sunbird-video-player *ngIf="playerConfig.metadata.mimeType === 'video/mp4'" [playerConfig]="playerConfig"
```
HarishGangula, line 37: mimeType may not be only mp4 it can be others also check with KP team once
we should use all video types other than youtube

HarishGangula (changes requested):
check all places where the player is used even in the workspace module and add it, 
Hint: check new pdf player implementation by searching in project

HarishGangula (changes requested):
You need to add these changes in all the places like it is missing in the workspace module
**Hint:** Check new pdf player implementation locations in code base add video player there also by search in project

src/app/client/src/app/modules/player-helper/components/player/player.component.html
souravdey091: Done

HarishGangula (approved):

PR #5563, opened 2020-12-02
Issue #SB-21863 fix:  Secure forum endpoints

swetha-aggidevara:


src/app/routes/ssoRoutes.js
```diff
@@ -423,8 +423,8 @@ module.exports = (app) => {
     }
   });
 
-  app.all('/get/user/sessionId', (req, res) => {
-      res.send({id: getEncyptedQueryParams({userName: req.query.userName})})
+  app.all('/learner/get/user/sessionId', (req, res) => {
+      res.send({id: getEncyptedQueryParams({userName: req.session.userName})})
```
vinukumar-vs, line 427: This is not going to solve the problem, You have to take userId as input & compare with session userId. If it matches then send the userName else through error

vinukumar-vs: this way we will secure this endpoint. Other users can;t take someother users token.

vinukumar-vs (changes requested):

vinukumar-vs (approved):

PR #5569, opened 2020-12-02
Issue #SB-21863 rever: Revert "Issue #SB-21863 fix:  Secure forum endpoints"

vinukumar-vs:
Reverts Sunbird-Ed/SunbirdEd-portal#5563

### DIFF
--- a/src/app/routes/ssoRoutes.js
+++ b/src/app/routes/ssoRoutes.js
@@ -423,13 +423,8 @@ module.exports = (app) => {
     }
   });
 
-  app.all('/learner/get/user/sessionId/:userId', (req, res) => {
-      if (req.session.userId === req.params.userId) {
-        res.send({id: getEncyptedQueryParams({userName: req.session.userName})})
-      } else {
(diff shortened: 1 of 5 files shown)


src/app/routes/ssoRoutes.js
```diff
@@ -423,13 +423,8 @@ module.exports = (app) => {
...
   });
 
-  app.all('/learner/get/user/sessionId/:userId', (req, res) => {
-      if (req.session.userId === req.params.userId) {
-        res.send({id: getEncyptedQueryParams({userName: req.session.userName})})
-      } else {
-        throw 'unhandled exception while getting sessionID';
-      }
-
+  app.all('/get/user/sessionId', (req, res) => {
```
rajeevsathish, line 426: here we can use the .get only instood of .all
and how are we seeing if this is called by a loggedin user only

rajeevsathish (commented):
the test case is failing

rajeevsathish (approved):

PR #5603, opened 2020-12-09
Issue #DP-1414 Fix:Adopted textbooks should not be editable

prashantsingh03:


prashantsingh03:
PR raised again with some changes

PR #5627, opened 2020-12-11
Issue #SH-21192 fix: Language change event reverted

itsvick:


src/desktop/modules/routes/static.ts
```diff
@@ -88,6 +88,7 @@ export default (app, contentFilesPath, ecarsFolderPath ) => {
     locals.deviceProfileApi = "/api/v3/device/profile";
     locals.deviceApi = `${process.env.APP_BASE_URL}/api/`;
     locals.baseUrl = process.env.APP_BASE_URL;
```
HarishGangula, line 90: Alread we have base url why offlineDesktopAppDownloadUrl  is required

src/desktop/scripts/metadata.js
```diff
@@ -289,9 +289,9 @@ const copyAssetsAndUpdateFiles = async () => {
 }
 
 const init = async () => {
-    await fse.copyFile(path.join(__dirname, `..`, `..`, `desktop-assets/${process.env.offline_target_env}/appConfig.json`), `env.json`)
+    // await fse.copyFile(path.join(__dirname, `..`, `..`, `desktop-assets/${process.env.offline_target_env}/appConfig.json`), `env.json`)
```
HarishGangula, line 292: Why these lines are commented

HarishGangula (changes requested):

HarishGangula (approved):

PR #5641, opened 2020-12-15
Issue#SB-22101 Course creator is not able to select different certificate templates when multiple default templates are configured

ArtiSingh23:


vinukumar-vs:
@ArtiSingh23 Can you create the new PR again..

PR #5663, opened 2020-12-17
Issue #SB-21199 feat: Added CRUD for logged in users

itsvick:


src/desktop/OpenRAP/sdks/UserSDK.ts
```diff
@@ -69,16 +68,85 @@ export class UserSDK {
...
+        code: "BAD_REQUEST",
+        status: 400,
+        message: `_id is mandatory to update user`
+      }
+    }
+    return this.dbSDK.delete(USER_DB, id);
+  }
+
+  public async getUserToken() {
+    return await this.settingSDK.get('oauth_token');
```
HarishGangula, line 124: This is not oauth_token it will be a session you will set and get and token should be part of user object only and should be omited while sending user object to frontend

HarishGangula (changes requested):

HarishGangula (approved):

PR #5674, opened 2020-12-18
Issue #SB-21199 feat: Managed user fetch/store implementation

itsvick:


src/desktop/modules/controllers/authController.ts
```diff
@@ -51,8 +51,21 @@ export default class AuthController {
             }
 
             const user = await permissionsHelper.getUser(userAuthDetails);
+            const managedUsers: any = await permissionsHelper.getManagedUsers(userAuthDetails);
```
HarishGangula, line 54: SDK should not have dependencies from outside of OpenRAP instead send all the users (main user + Managed user at once) to insert it

src/desktop/modules/helper/permissionsHelper.ts
```diff
@@ -73,7 +70,31 @@ const PERMISSIONS_HELPER = {
       logger.error("Error while getting user", error);
       throw { message: `User read failed with ${error}`, status: error.code || 500 }
     }
-  }
+  },
+  // Fetch user data from server
+  async getManagedUsers(managedByUser: { access_token: string, userId: string }): Promise<ILoggedInUser> {
```
HarishGangula, line 75: In this method, the error messages should be relevant to the context which becomes hard to debug in the field

HarishGangula (changes requested):

HarishGangula (approved):

PR #5709, opened 2020-12-24
Issue #SB-22197 fix: fixed QR code global search functionality

Nik720:


src/desktop/modules/controllers/content/content.ts
```diff
@@ -73,7 +73,17 @@ export default class Content {
...
         logger.debug(`ReqId = "${reqId}": Find the contents in ContentDb with the given filters`)
-        return this.databaseSdk.find('content', dbFilters);
+        const dbResult = await this.databaseSdk.find('content', dbFilters);
+        let QRresult = [];
+        if (_.get(filters, 'query')) {
+            QRresult  = await this.searchForDialCodeContent(_.get(filters, 'query'));
+        } 
+        if(QRresult.length > 0) {
+            _.forEach(QRresult, (obj, index) => {
+                dbResult.docs.push(obj);
```
HarishGangula, line 83: Here there is possiblility of duplicate content so make unique by identifier

HarishGangula (changes requested):

src/desktop/modules/controllers/content/content.ts
Nik720: done

PR #5729, opened 2020-12-31
Issue #SB-21199 feat: Used same deviceId from desktop backend for desktop app

itsvick:


src/app/client/src/app/app.component.ts
```diff
@@ -493,6 +493,9 @@ export class AppComponent implements OnInit, OnDestroy {
    */
   public setDeviceId(): Observable<string> {
       return new Observable(observer => this.telemetryService.getDeviceId((deviceId, components, version) => {
+        if (this.utilService.isDesktopApp) {
```
HarishGangula, line 496: There should be else condition otherwise it  overrides with fingerPrint js device id

HarishGangula (changes requested):

HarishGangula (approved):

PR #5735, opened 2021-01-05
Issue #SB-21642 feat: Implementation of Managed user switch

itsvick:


src/desktop/modules/routes/auth.ts
```diff
@@ -138,6 +139,27 @@ export default (app, proxyURL) => {
...
+      logger.info({ msg: 'switch user rejected missing userID' });
+      res.status(400).send(Response.error("api.user.switch", 400, "failed to switch user", "BAD_REQUEST"));
+    }
+
+    try {
+      const sessionData = { userId: req.params.userId };
+      await userSDK.setUserSession(sessionData);
+      const result = {
+        response: "Success",
+        userSid: uuidv1()
```
HarishGangula, line 154: is the same sessionId is not stored in DB ?

HarishGangula (changes requested):

src/desktop/modules/routes/auth.ts
itsvick: No, its not stored

HarishGangula: sessionId should be stored in DB and used in server side telemetry as well

HarishGangula (approved):

PR #5737, opened 2021-01-05
propagation of trace Id

swetha-aggidevara:


src/app/client/src/app/modules/core/services/data/data.service.ts
```diff
@@ -187,7 +187,9 @@ export class DataService {
       'X-msgid': _uuid,
       'X-Request-ID': _uuid,
       'X-App-Version': this.appVersion,
-      'X-Session-ID': DataService.sessionId
+      'X-Session-ID': DataService.sessionId,
+      'X-Trace-ID': _uuid,
+      'X-Trace-Enabled': true
```
vinukumar-vs, line 192: Use this as an environment variable.

src/app/client/src/app/service/CsLibInitializer/cs-lib-initializer.service.ts
```diff
@@ -48,7 +49,23 @@ export class CsLibInitializerService {
                 certRegistrationApiPath: '/learner/certreg/v2/certs'
               }
           }
-      });
+      },
+      null,
+      new class implements CsClientStorage {
```
vinukumar-vs, line 54: Why are we using this?

vinukumar-vs (changes requested):

src/app/client/src/app/service/CsLibInitializer/cs-lib-initializer.service.ts
swetha-aggidevara: This is for CSL related TRACE-ID

vinukumar-vs (approved):

PR #5744, opened 2021-01-05
Issue #SB-22183 fix: Addig user token in session object

VenkatKadiveti:


src/app/routes/discussionsForum.js
```diff
@@ -149,10 +149,15 @@ function checkEmail() {
             }
     }
 }
-
+function addHeaders() {
+    return function (proxyReqOpts, srcReq) {
```
vinukumar-vs, line 153: Call 
proxyUtils.decorateRequestHeaders(discussions_middleware),
and then ovviride Authorization with below code.

vinukumar-vs (changes requested):

PR #5811, opened 2021-01-12
Issue #SB-21199 fix: added logged in user id as actor and device id for non logged in user

Nik720:


HarishGangula (approved):

src/desktop/OpenRAP/sdks/UserSDK.ts
```diff
@@ -148,7 +148,11 @@ export class UserSDK {
   }
 
   public async getUserSession() {
```
HarishGangula, line 150: No need to handle like this return the promise and handle at other ends with catch

HarishGangula (changes requested):

HarishGangula (approved):

PR #5903, opened 2021-01-25
Issue #SB-22726 fix: fixed desktop mac build related issue

Nik720:


src/desktop/OpenRAP/managers/TelemetryManager.ts
```diff
@@ -88,7 +88,12 @@ export class TelemetryManager {
 
   async batchJob() {
     try {
-      const loggedInUser = await this.userSDK.getUserSession();
+      let loggedInUser;
+      this.userSDK.getUserSession().then(data => {
```
HarishGangula, line 92: This won't work
it can be
` let loggedInUser = ;
loggedInUser = await this.userSDK.getUserSession().catch(err => logger.error("<error message>"));
`

HarishGangula (changes requested):

HarishGangula (approved):

PR #5999, opened 2021-02-11
Issue#SB-22458 Discussion Forum: Theming support for library add DF t…

ArtiSingh23:
…heme css inside portal theme file


vinukumar-vs (approved):

src/app/client/src/assets/styles/themes.scss
```diff
@@ -1194,4 +1194,77 @@ html[layout='joy'] {
...
+    border-radius: var(--sbt-bradius-md);
+    margin-right: 1.5rem;
+
+    @media (max-width: 768px) {
+      margin-right: 0;
+      padding: 1.5rem;
+    }
+  }
+
+  .df-filter-option {
```
vinukumar-vs, line 1241: We should not add the class names here. It should be with viable names only

src/app/client/src/assets/styles/themes.scss
```diff
@@ -1194,4 +1194,77 @@ html[layout='joy'] {
...
+}
+
+html[layout='joy'] {
+  --df-primary-bg: #FFD954;
+  --df-gray-bg: #E9E8D9;
+  --sbt-bradius: 0.125rem;
+  --sbt-bradius-xs: calc(var(--sbt-bradius) * 4);  /* 8 value */
+  --sbt-bradius-sm: calc(var(--sbt-bradius) * 8);  /* 16 value */
+  --sbt-bradius-md: calc(var(--sbt-bradius) * 12);  /* 24 value */
+  --sbt-bradius-lg: calc(var(--sbt-bradius) * 16);  /* 32 value */
```
vinukumar-vs, line 1206: Till here the code is right. Below there are too many classes are added. We shouldn't add any classes here.

vinukumar-vs (changes requested):

PR #6077, opened 2021-02-23
Issue #SB-23210 feat: Added system warning component

itsvick:


src/app/client/src/app/modules/core/services/public-data/public-data.service.ts
```diff
@@ -1,4 +1,4 @@
-import { ConfigService } from '@sunbird/shared';
+import { ConfigService } from '../../../shared/services/config/config.service';
```
HarishGangula, line 1: Why this import is changed

src/app/client/src/app/modules/public/module/offline/services/system-info/system-info.service.ts
```diff
@@ -1,6 +1,6 @@
 import { Injectable } from '@angular/core';
-import { PublicDataService } from '@sunbird/core';
-import { ConfigService } from '@sunbird/shared';
+import { PublicDataService } from '../../../../../core/services/public-data/public-data.service';
```
HarishGangula, line 2: Here also imports are changed

HarishGangula (changes requested):

src/app/client/src/app/modules/public/module/offline/services/system-info/system-info.service.ts
itsvick: There was a circular dependency, due to this, as we pointing to index.ts where all the services were included

HarishGangula (approved):

PR #6258, opened 2021-03-22
Issue #SB-23638 fix: Added close button functionality in DF library

sudip-mukherjee:


src/app/client/src/app/modules/groups/components/group-header/group-header.component.ts
```diff
@@ -223,6 +223,7 @@ export class GroupHeaderComponent implements OnInit, OnDestroy {
         type: 'Group'
       }
     ];
+    this.discussionTelemetryService.parentRouteUrl = window.location.href;
```
vinukumar-vs, line 226: Why don't we use NavigationHelperService?

discussionTelemetryService should have logic specific to only telemetry

vinukumar-vs (changes requested):

src/app/client/src/app/modules/groups/components/group-header/group-header.component.ts
sudip-mukherjee: Done

vinukumar-vs (approved):

PR #6277, opened 2021-03-25
Issue #SB-21198 fix: Show download status for imported content on course page and updated button text

itsvick:


src/app/client/src/app/modules/public/module/offline/services/content-manager/content-manager.service.ts
```diff
@@ -36,7 +36,6 @@ export class ContentManagerService {
   updateContentDownloadStatus(contentDownloadList) {
     this.contentDownloadStatus = {};
     _.forEach(contentDownloadList, content => {
-      if (content.addedUsing === 'download') {
```
HarishGangula, line 39: WHy we remove this condition ?

HarishGangula (changes requested):

src/app/client/src/app/modules/public/module/offline/services/content-manager/content-manager.service.ts
itsvick: Because it saves the status of downloaded content only, not for imported content. So, download status is not getting shown on the course toc page when we import the content.

HarishGangula (approved):

PR #6354, opened 2021-04-09
Issue #SB-23815 feat: Show year of birth popup for all user except SSO user and new users

Ajoymaity:


src/app/client/src/app/app.component.html
```diff
@@ -71,6 +71,8 @@
       [userFeedData]="userFeed"> </app-validate-teacher-identifier-popup>
     <app-user-onboarding (close)="onLocationSubmit()" *ngIf="showUserTypePopup || !isLocationConfirmed"
       [isCustodianOrgUser]="isCustodianOrgUser" [deviceProfile]="deviceProfile"></app-user-onboarding>
+    
+    <app-year-of-birth *ngIf="isCustodianOrgUser && usersProfile && !usersProfile?.dob && (usersProfile?.userLocations?.length>0 || showYearOfBirthPopup)"></app-year-of-birth>
```
rajeevsathish, line 75: is this only needed for CustodianOrgUser not the state users?

rajeevsathish (changes requested):

src/app/client/src/app/app.component.html
Ajoymaity: yes, this is need for Custodian users only because,
Assumption : when a user is signing up using SSO , they will be always above 18

rajeevsathish (approved):

PR #6641, opened 2021-06-02
Issue #SB-23152 fix: Update UI and state management issue

rajeshkumaravel:
- Issue #SB-23152 fix: Update UI and state management issue


rajeshkumaravel:
Ref: #6643

PR #6685, opened 2021-06-09
Release 4.0.0

kiranharidas187:


src/app/client/src/app/modules/observation/guards/ml/ml.guard.ts
```diff
@@ -24,6 +24,10 @@ export class MlGuard implements CanActivate {
         return true
       } else {
         this.toasterService.error(_.get(this.resourceService, 'messages.stmsg.m0145'));
+        let queryParam = {
```
rajeevsathish, line 27: if it is not changing please make it a const and as it is object it can be const
and you can change its attributes

rajeevsathish (changes requested):

rajeevsathish (approved):

PR #6697, opened 2021-06-10
Issue #SB-24324 fix: Portal backend error log

kamaleeswary:


src/app/helpers/telemetryHelper.js
```diff
@@ -529,5 +529,23 @@ module.exports = {
       type: 'UserSession'
     });
     return cdata;
+  },
+
+/**
+ * This function used to generate api_access log event
+ */
+  getTelemetryAPISuceessData: function(result, req, uri) {
```
vinukumar-vs, line 537: We dont need new method here. We can use the existing log method itself by passing right EDATA.

vinukumar-vs (changes requested):

src/app/helpers/telemetryHelper.js
kamaleeswary: in this method, I m preparing the req data for the existing function.

src/app/libs/sb_telemetry_util/telemetryService.js
```diff
@@ -320,22 +327,21 @@ telemetryService.prototype.generateApiCallLogEvent = function (data) {
...
   try {
-  if ( (data.responseCode !== 'OK' && data.responseCode !== 200) || res.statusCode !== 200) {
-  const result = data.params;
-  if (result) {
-  const edata = {
-    err: data.responseCode  || res.statusCode,
-    errtype: result.err || res.statusMessage,
-    requestid:  result.resmsgid || 'null',
-    errmsg: result.errmsg  || 'null',
+    const result = data.params;
```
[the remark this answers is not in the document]
kamaleeswary: we don't need this 'if ( (data.responseCode !== 'OK' && data.responseCode !== 200) || res.statusCode !== 200) '
check here, only the error events should call this function then there is no need to check the statuscode. the rest inside the data i just changed the param name to match the standard that's it.

src/app/routes/groupRoutes.js
```diff
@@ -42,26 +42,29 @@ function proxyObj() {
...
             } catch (err) {
-                const uri = 'learner/group'
-                const context = {
-                    env: telemtryEventConfig.URL[uri].env || 'group'
-                }
-                const option = telemetry.getTelemetryAPIError(JSON.parse(resData),proxyRes, context);
-                if (option) {
-                telemetryHelper.logApiErrorEventV2(req, option);
-                }
-                logger.error({ msg: 'learner route : userResDecorator json parse error:', proxyResData })
```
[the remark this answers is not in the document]
kamaleeswary: catch block has only the code or any functional failure right. if any API event is failed, then it will be in try block only, and there we have one more condition right to check the status…if we don’t use the condition also it will not go to catch block right?

balakrishna10 (approved):

src/app/helpers/telemetryHelper.js
```diff
@@ -131,17 +131,18 @@ module.exports = {
...
       { 'category': apiConfig && apiConfig.category },
-      { 'url': apiConfig && apiConfig.url },
+      { 'url': options.path || apiConfig && apiConfig.url },
       { 'duration': Date.now() - new Date(options.headers.ts) },
       { 'status': statusCode },
       { 'protocol': 'https' },
       { 'method': options.method },
-      { 'req': options.body }
+      { 'traceid': traceid },
+      { 'req': options.body },
```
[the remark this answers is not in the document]
kamaleeswary: this is req.body i didn't add...its already defined in this method...

src/app/helpers/telemetryHelper.js
```diff
@@ -285,20 +289,30 @@ module.exports = {
       })
     }
   },
-  logApiErrorEventV2: function (req, options) {
+  logApiErrorEventV2: function (req, options, exception = false) {
```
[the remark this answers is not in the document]
kamaleeswary: ok

src/app/helpers/telemetryHelper.js
```diff
@@ -131,17 +131,17 @@ module.exports = {
...
       { 'title': apiConfig && apiConfig.title },
       { 'category': apiConfig && apiConfig.category },
-      { 'url': apiConfig && apiConfig.url },
+      { 'url': options.path || apiConfig && apiConfig.url },
       { 'duration': Date.now() - new Date(options.headers.ts) },
       { 'status': statusCode },
       { 'protocol': 'https' },
       { 'method': options.method },
-      { 'req': options.body }
+      { 'traceid': traceid }
```
VenkatKadiveti, line 144: is req not required here ?
why did you remove ?

src/app/helpers/telemetryHelper.js
```diff
@@ -217,15 +217,18 @@ module.exports = {
...
     let object = {}
     let params = []
-    if (req.options) {
-      params = this.getParamsData(req.options, req.statusCode, req.resp, req.uri)
+    if (req.reqObj) {
+      params = this.getParamsData(req.reqObj, req.statusCode, req.resp, uri)
     }
-    const edata = telemetry.logEventData('api_access', 'INFO', apiConfig.message, params)
+
+    const edata = telemetry.logEventData('api_access', 'INFO', apiConfig.message,JSON.stringify(params))
```
VenkatKadiveti, line 230: use constants instead of using hard coded strings

src/app/helpers/telemetryHelper.js
```diff
@@ -54,7 +54,7 @@ module.exports = {
     context.did = req.session.deviceId
     context.rollup = telemetry.getRollUpData(dims)
     const actor = telemetry.getActorData(req.session.userId, 'user')
-   console.log('logging session start event', context.did);
+    console.log('logging session start event', context.did);
```
VenkatKadiveti, line 57: we have logger instead of using console.log we can use 
logger.info({ msg: 'string' });

balakrishna10 (approved):

balakrishna10 (commented):
Please check Venkats review.

balakrishna10 (commented):
Please check Venkats review.

src/app/helpers/telemetryHelper.js
kamaleeswary: this one I haven't add, this line is already there, only lint errors are fixed.

src/app/helpers/telemetryHelper.js
kamaleeswary: this method is for entry log , in the previous review vinu said , we no need to pass the req body in this method.so I removed it here

src/app/helpers/telemetryHelper.js
kamaleeswary: ok

VenkatKadiveti (approved):

vinukumar-vs (approved):

PR #6837, opened 2021-07-07
Issue #SB-24116 fix: DF looping issue on navigation back

balakrishna10:


src/app/client/src/app/modules/shared/services/navigation-helper/navigation-helper.service.ts
```diff
@@ -234,4 +234,15 @@ export class NavigationHelperService {
       this.router.navigate(['/resources']);
     }
   }
+
+  /* Used for removing DF routes and duplicate routes from history which will cause looping */
+  removeDfRoutes() {
+    this._history = this._history.filter(h => !(h.url.includes('discussion-forum')))
```
vinukumar-vs, line 240: Navigation helper service should not have any code specific to any component(like her we are using 'discussion-forum' string). We should not add any checks like this in this server.

Find the actual root cause for failure.

src/app/client/src/app/modules/groups/services/groups/groups.service.ts
```diff
@@ -178,6 +178,9 @@ getActivity(groupId, activity, mergeGroup, leafNodesCount?) {
     if (this.navigationhelperService['_history'].length <= 1) {
       this.router.navigate([MY_GROUPS]);
     } else {
+      if(this.navigationhelperService['_history'].find(n => n.url.indexOf('discussion-forum') != -1)){
```
vinukumar-vs, line 181: We should not have these kind of checks.

vinukumar-vs (changes requested):

sudip-mukherjee (approved):

vinukumar-vs (approved):

PR #6963, opened 2021-07-22
Issue #SB-21560 feat: Front End Exception logs for User and Org-learner

kamaleeswary:


src/app/client/src/app/modules/groups/services/groups/groups.service.ts
```diff
@@ -373,4 +373,32 @@ getActivity(groupId, activity, mergeGroup, leafNodesCount?) {
   getDashletData(courseHeirarchyData, aggData) {
     return this.groupCservice.activityService.getDataForDashlets(courseHeirarchyData, aggData);
   }
+
+  // Temporary fix to get telemetry contexrt for group exception logs 
+
+  /**
+   * generate the telemetry context data by using router details
+   */
+  getTelemetryContext() {
```
vinukumar-vs, line 382: Too much of code, that to lot fo assumptions specific to activity & group. 
not happy with the logic being used even though refactoring has to be done for this.

src/app/client/src/app/modules/telemetry/interfaces/telemetry.ts
```diff
@@ -49,6 +49,7 @@ export interface IErrorEventData {
   'err': string;
   'errtype': string;
   'stacktrace': string;
+  'traceid': string;
```
vinukumar-vs, line 52: This should be options. Otherwise all the existing code will break.

vinukumar-vs (approved):

PR #6978, opened 2021-07-26
Issue #SB-25859 fix: updated user slug by uid

kamaleeswary:


src/app/client/src/app/modules/discussion/components/access-discussion/access-discussion.component.spec.data.ts
```diff
@@ -37,7 +37,7 @@ export const MockResponseData = {
       },
       emitData: {
         'forumIds': '[6]',
-        'userName': 'cctn1350'
+        'usrId': 1
```
vinukumar-vs, line 40: typo error

vinukumar-vs (changes requested):

src/app/client/src/app/modules/discussion/components/access-discussion/access-discussion.component.spec.data.ts
kamaleeswary: it's correct only right, it's in number.

VenkatKadiveti: userslug is string type but the userID is integer only

vinukumar-vs (approved):

PR #7046, opened 2021-08-10
Release 4.2.0

kiranharidas187:


src/app/client/src/app/modules/explore-page/components/explore-page/explore-page.component.ts
```diff
@@ -528,6 +542,46 @@ export class ExplorePageComponent implements OnInit, OnDestroy, AfterViewInit {
...
+                        data[currentRole][this.userType]) {
+                        this.showTargetedCategory = true
+                        this.targetedCategory = data[currentRole][this.userType];
+                        this.targetedCategorytheme = {
+                            "iconBgColor": "rgba(255,255,255,1)",
+                            "pillBgColor": "rgba(255,255,255,1)"
+                        }
+                    }
+                    else {
+                        this.showTargetedCategory = false
```
rajeevsathish, line 576: make this by default as false then the else will not be needed

rajeevsathish (changes requested):

src/app/client/src/app/modules/core/components/main-header/main-header.component.ts
```diff
@@ -188,19 +189,56 @@ export class MainHeaderComponent implements OnInit, OnDestroy {
...
         }
       }
     });
+  }
 
+  getFormConfigs() {
+    this.observationUtilService.browseByCategoryForm()
+      .then((data: any) => {
+        let currentBoard;
+        Array.isArray(this.userPreference.framework.id) ? (currentBoard=this.userPreference.framework.id[0]) : (currentBoard=this.userPreference.framework.id)
```
rajeevsathish, line 214: if the object like this.userPreference or the framework is undefined then this will break the entire flow with out checking if we use . operator on the object

rajeevsathish (changes requested):

rajeevsathish (approved):

PR #7171, opened 2021-09-13
Program Dashboard 

rakeshSgr:


src/app/client/src/app/modules/core/components/main-header/main-header.component.html
```diff
@@ -404,6 +404,14 @@ <h4 class="mb-4 text-capitalize">{{showingResult}}</h4>
               </svg></i>
             {{resourceService.frmelmnts?.lnk?.dashboard}}
           </li>
+          <li class="list-item" *ngIf='!isDesktopApp && permissionService.permissionAvailable' appPermission
+          [permission]='programDashboardRole' (click)="toggleSideMenu(false)" routerLink="program">
+          <i class="icon-svg icon-svg--sm icon-dashboard mr-8"><svg class="icon icon-svg--primary">
+              <use xlink:href="./assets/images/sprite.svg#dashboard"></use>
+            </svg></i>
+            Program Dashboard
```
rajeevsathish, line 412: please do not hard code the value

rajeevsathish: for this and the dashboard the image is the same?

rajeevsathish (changes requested):

src/app/client/src/app/modules/core/components/main-header/main-header.component.html
rakeshSgr: I have fixed hard code value, Image is same only, program-dashboard can be accessible only with role of PROGRAM_MANAGER and PROGRAM_DESIGNER

rajeevsathish (approved):

rakeshSgr:
Code coverage 70% in local
image (1)

PR #7181, opened 2021-09-14
#SB-25482 chore: updated sunbird collection package version

santoshilimi:


src/app/client/yarn.lock
```diff
@@ -1490,7 +1490,7 @@
     whatwg-fetch "^3.1.0"
```
vaivk369, line 1491: @santoshilimi do you want to submit this file?

PR #7512, opened 2021-11-29
Issue #SB-27106 Remove activity button in popup is not get highighted post select remove activity from 3 dot.

ArtiSingh23:
…select remove acitivity from 3 dot

### DIFF
--- a/src/app/client/src/app/modules/groups/components/activity/activity-list/activity-list.component.html
+++ b/src/app/client/src/app/modules/groups/components/activity/activity-list/activity-list.component.html
@@ -38,17 +38,21 @@
 <app-add-activity *ngIf="!showLoader && ((activityList | json) === '{}')"></app-add-activity>
 
 <!-- Remove Activity from the group -->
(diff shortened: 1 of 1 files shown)


src/app/client/src/app/modules/groups/components/activity/activity-list/activity-list.component.html
```diff
@@ -38,17 +38,21 @@
...
-<div class="sb-modal-content">
-  <p> {{resourceService?.messages?.imsg?.removeActivityWarning}}</p>
-</div>
-<div class="sb-modal-actions">
-  <button class="sb-btn sb-btn-normal sb-btn-error" tabindex="0"
-    (click)="removeActivity()">{{resourceService?.frmelmnts?.btn?.remove}}</button>
-</div>
-
-</sui-modal>
+<app-modal-wrapper *ngIf="showModal" #modal [config]="{disableClose: true, size: 'small'}">
```
vinukumar-vs, line 41: @ArtiSingh23 @VenkatKadiveti What about (dismissed)="toggleModal()" option. Is that use-case is working fine?

vinukumar-vs (approved):

PR #7521, opened 2021-11-30
Issue #SB-27628 fix: In Groups the user not able to add the activity like course & trackable collection.

kamaleeswary:


src/app/client/src/app/modules/groups/components/activity/activity-search/activity-search.component.ts
```diff
@@ -239,6 +239,7 @@ export class ActivitySearchComponent implements OnInit, OnDestroy {
     /* istanbul ignore else */
     if (_.get(this.queryParams, 'key')) {
       option.query = this.queryParams.key;
+      option.filters = _.pick(filters, 'primaryCategory', 'objectType', 'status');
```
vinukumar-vs, line 242: Why this change is added now. WHat is the root cause for this?
cc: @balakrishna10

vinukumar-vs (changes requested):
Identify the root cause why it is failig now.

PR #7703, opened 2022-01-24
Release 4.7.0

rajeevsathish:


rajeshkumaravel (approved):

rajeevsathish:
the sonar is looking for a coverage for a new line of if condition which cannot be written test case for so cannot cover that so can you please merge this PR

PR #7782, opened 2022-02-15
Issue #SB-28659:fix: Asterisk symbol added for board

kiranharidas187:


rajeevsathish (changes requested):
please do not remove the if conditions
but add the required as yes in the form

rajeevsathish (approved):

PR #7822, opened 2022-02-22
Issue #Edportal : fix : package.json updated

pradeepkumarcm96:


rajeevsathish (commented):
where is the yarn.lock file

PR #7918, opened 2022-03-23
fix: SB-29293, SB-29300 and SB-29225

manojSRawat:


src/app/resourcebundles/data/consumption/en.properties
```diff
@@ -2079,6 +2079,7 @@ frmelmnts.lbl.enterPurpose = Enter purpose
 frmelmnts.lbl.enterMessage = Enter message
 frmelmnts.lbl.uploadOdkForm = Upload ODK Form (.xml)
 frmelmnts.lbl.sampleOdkForm = Sample ODK Form (.xlsx)
+frmelmnts.lbl.browseXmlFile = Browse XML File
```
rajeevsathish, line 2082: If it is used specific to uci please use ucibrouseXmlFile so that if some one tries to change the value they will have a idea that this is used by other team

rajeevsathish (changes requested):
If labels are used specific to uci please use ucibrouseXmlFile so that if some one tries to change the value they will have a idea that this is used by other team

src/app/resourcebundles/data/consumption/en.properties
manojSRawat: No these are not specific to UCI, anyone can use this. I added a UCI comment to categorize it.

rajeevsathish (approved):

PR #8054, opened 2022-05-25
#SB-26905 feat: updated video player version

santoshilimi:


santoshilimi:
Updated in different PR

PR #8275, opened 2022-09-07
fix: SB-30832 and SB-30816

manojSRawat:


vinukumar-vs (dismissed):

src/app/resourcebundles/data/consumption/hi.properties
```diff
@@ -1990,4 +1990,6 @@ frmelmnts.lbl.of = का
 frmelmnts.lbl.boardsTest=Board {frameworkCategory1}
 frmelmnts.lbl.mediumTest=Medium {frameworkCategory2}
 frmelmnts.lbl.gradeTest=Grade {frameworkCategory3}
-frmelmnts.lbl.boardsTestWithText=बोर्ड {frameworkCategory3}
\ No newline at end of file
+frmelmnts.lbl.boardsTestWithText=बोर्ड {frameworkCategory3}
```
rajeevsathish, line 1993: with out the JSON file the properties file will not work

rajeevsathish (changes requested):

src/app/resourcebundles/data/consumption/hi.properties
manojSRawat: we only added the last line `frmelmnts.lbl.questionSets = प्रश्न सेट`. Above mentioned line is already there

tarang27: @rajeevsathish can you please merge this PR. As Manoj mentioned we have not added the line on which you have shared the comment. 
Thanks

rajeevsathish (changes requested):
If only hi.properties has change then please push only hi.json not all

PR #8363, opened 2022-11-23
Issue #RRHE-52 fix:Status filter deselection of all values case

5Amogh:


src/app/client/src/app/modules/program-dashboard/components/program-datasets/program-datasets.component.ts
```diff
@@ -592,8 +592,12 @@ export class DatasetsComponent implements OnInit, OnDestroy {
         }else{
           this.configuredFilters[reference[0]] = value[0] as number -1;
         }
-      }else{
-        this.configuredFilters[reference[0]] = value[0]
+      }else if($event.controlType === 'multi-select'){
```
rajeevsathish, line 595: what is the case if the controlType is not equal to multi-select? or number

5Amogh: Currently the PD filter component (from where this data is getting emitted)has only number and multi-select control type filters, going forward if we get requirements for other control types that will be developed in the PD filter component and related logic will be included as well.

rajeevsathish (approved):

PR #8370, opened 2022-11-25
Issue #SB-0000 Updated:common-consumption to Ng-v11

princegupta1131:
updated : common-consumption v11 & content-section v11 to portal


src/app/client/package.json
```diff
@@ -57,7 +57,7 @@
     "@project-sunbird/ckeditor-build-classic": "4.1.3",
     "@project-sunbird/ckeditor-build-font": "^1.0.8",
     "@project-sunbird/client-services": "4.8.13",
-    "@project-sunbird/common-consumption-v9": "^4.10.3",
+    "@project-sunbird/common-consumption-v11": "4.12.0",
```
rajeevsathish, line 60: as I remember we said we will remove naming like _v11 and only use commin-consumption with the version we will come to know which angular version it will use? did it change

PR #8721, opened 2023-05-25
ED-1945  feat : implemented Copy Question Set feature in ED-Portal

princegupta1131:
# SunbirdEd - Portal
---

### Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.
---

#### Example
- [x] Applicable
- [ ] Not applicable

### Type of change

Please choose appropriate options.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Enhancement (additive changes to improve performance)
- [ ] This change requires a documentation update

### Checklist:

- [x] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/src/app/client/src/app/modules/core/services/copy-content/copy-content.service.ts
+++ b/src/app/client/src/app/modules/core/services/copy-content/copy-content.service.ts
@@ -67,9 +67,10 @@ export class CopyContentService {
               url: urlPath + '/' + contentData.identifier,
               data: param
             };
+            let reqParms =  _.get(contentData,'mimeType') === 'application/vnd.sunbird.questionset' ? param.request.questionset : param.request.content;
             return this.contentService.post(option).pipe(map((response: ServerResponse) => {
               _.forEach(response.result.node_id, (value) => {
-                this.redirectToEditor(param.request.content, value);
+                this.redirectToEditor(reqParms, value);
               });
               return response;
             }));
@@ -137,6 +138,7 @@ export class CopyContentService {
             name: 'Copy of ' + contentData.name,
             createdFor: userData.organisationIds,
             createdBy: userData.userId,
+            mimeType: contentData.mimeType,
             
           }
         }
--- a/src/app/client/src/app/modules/core/services/copy-content/copy-content.service.ts
+++ b/src/app/client/src/app/modules/core/services/copy-content/copy-content.service.ts
@@ -7,7 +7,6 @@ import { UserService } from '../../services/user/user.service';
 import { ContentService } from './../content/content.service';
 import { FrameworkService } from './../framework/framework.service';
 import { of } from 'rxjs';
-import { PublicDataService } from './../public-data/public-data.service';
 
 /**
  * Service to copy content
@@ -47,7 +46,7 @@ export class CopyContentService {
    * @param {ContentService} contentService ContentService reference
    */
   constructor(config: ConfigService, router: Router, userService: UserService, contentService: ContentService,
-    frameworkService: FrameworkService, public publicDataService: PublicDataService) {
+    frameworkService: FrameworkService) {
     this.config = config;
     this.router = router;
     this.userService = userService;
--- a/src/app/client/src/app/modules/core/services/copy-content/copy-content.service.ts
+++ b/src/app/client/src/app/modules/core/services/copy-content/copy-content.service.ts
@@ -190,7 +190,7 @@ export class CopyContentService {
       }
     } else if (contentData.mimeType === 'application/vnd.ekstep.ecml-archive') {
       url = `/workspace/content/edit/content/${copiedIdentifier}/draft/${contentData.framework}/Draft`;
-    } else if (contentData.mimeType === 'application/vnd.sunbird.questionset') {
+    } else if (_.get(contentData,'mimeType') === 'application/vnd.sunbird.questionset') {
       url = `/workspace/edit/QuestionSet/${copiedIdentifier}/allcontent/Draft`;
     }
     else {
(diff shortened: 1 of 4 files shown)


src/app/routes/contentRoutes.js
```diff
@@ -82,4 +82,27 @@ module.exports = (app) => {
                 }
             })
         )
+    app.post('/content/questionset/v1/copy/:id',
```
vinukumar-vs, line 85: Why did we wrote separate method for this? Can't we use the existing QuestionSet or ContentCopy methods itself, rather than writing different for each endpoint.

What is the differentiating factor to write separate method?

src/app/client/src/app/modules/core/services/copy-content/copy-content.service.ts
```diff
@@ -115,36 +116,52 @@ export class CopyContentService {
...
-          description: contentData.description,
-          code: contentData.code + '.copy',
-          creator: creator,
-          createdFor: userData.organisationIds,
-          createdBy: userData.userId,
-          organisation: _.uniq(this.userService.orgNames),
-          framework: '',
-          mimeType: contentData.mimeType,
-          contentType: contentData.contentType
+    if (_.get(contentData,'mimeType') === 'application/vnd.sunbird.questionset') {
```
vinukumar-vs, line 134: Why are we written separate if else  for questionset. 
Will the questionset copy fails if we send the same request that we are passing?

vinukumar-vs (changes requested):

src/app/client/src/app/modules/core/services/copy-content/copy-content.service.ts
princegupta1131: Request metadata is differ, 
 **question set request body** ::  req = {request: { questionset: {... } }  and for question-set content type metadata is not required 
**content request body**  req = { request: { content: { ...}} .

src/app/helpers/whitelistApis.js
```diff
@@ -2046,7 +2053,8 @@ const API_LIST = {
     '/uci/admin/v1/forms/upload',
     '/kendra/user-extension/mlcore/v1/solutions/:id',
     '/assessment/programs/mlcore/v1/join/:id',
-    '/assessment/users/mlcore/v1/solutions/:id'
+    '/assessment/users/mlcore/v1/solutions/:id',
+    '/questionset/v1/copy/:do_id'
```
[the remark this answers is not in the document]
princegupta1131: ok

src/app/helpers/whitelistApis.js
```diff
@@ -1694,6 +1694,13 @@ const API_LIST = {
         ROLE.CONTENT_CREATOR
       ]
     },
+    '/questionset/v1/copy/:do_id': {
```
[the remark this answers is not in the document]
princegupta1131: ok

src/app/client/src/app/modules/core/services/copy-content/copy-content.service.ts
```diff
@@ -174,7 +191,10 @@ export class CopyContentService {
       }
     } else if (contentData.mimeType === 'application/vnd.ekstep.ecml-archive') {
       url = `/workspace/content/edit/content/${copiedIdentifier}/draft/${contentData.framework}/Draft`;
-    } else {
+    } else if (_.get(contentData,'mimeType') === 'application/vnd.sunbird.questionset') {
```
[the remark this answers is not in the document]
princegupta1131: let me try this

src/app/client/src/app/modules/core/services/copy-content/copy-content.service.ts
vinukumar-vs: The same properties present in the Content request as well right. Can't I create the content with same properties of Question-Set?

src/app/routes/contentRoutes.js
rajeshkumaravel: @vinukumar-vs 
I see that content route is defined with wildcard and it is using `app.all`, as we have made practice not to use wild cards or `app.all` post **API whitelisting implementation**, as this is not best / recommended practice.
So, we are expecting request method type and appropriate definition for any new route being introduced in portal backend.
The changes looks good to me
CC: @rajeevsathish @princegupta1131

src/app/client/src/app/modules/core/services/copy-content/copy-content.service.ts
princegupta1131: @vinukumar-vs  
I agree, but according to the backend implementation  -> http://941.qa.docs.sunbird.org/4.9.0/apis/questionapi//#operation/Import%20QuestionSet
backend is expecting **request.questionset** in request body so, I have implemented the same.

CC: @rajeevsathish @rajeshkumaravel

src/app/client/src/app/modules/core/services/copy-content/copy-content.service.ts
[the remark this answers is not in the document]
princegupta1131: @vinukumar-vs 
I did the analysis,  by solving one function problem. if we are importing the workspace module into the core module. it will be imported into two different files. which will be a bumping of the bundle size. it will solve only the if-else method problem. we are not using any other method from workspace modules in core modules. 
and also some code-level changes are required to call this workspace .service.ts
Required parameter **state**: unable to find this metadata on click of copy content/questionset
need to add two additional metadata to call this service [identifier,status] 

CC: @rajeevsathish @rajeshkumaravel

src/app/helpers/whitelistApis.js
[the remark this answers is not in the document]
princegupta1131: fixed

rajeshkumaravel (approved):

PR #8725, opened 2023-05-26
Issue #ED-94 fix: chart config fix for program dashboard as required by sb-dashlet

5Amogh:


src/app/client/package.json
```diff
@@ -77,7 +77,7 @@
     "@project-sunbird/telemetry-sdk": "0.0.29",
     "@project-sunbird/web-extensions": "6.0.0",
     "@samagra-x/uci-console": "6.0.1",
-    "@shikshalokam/sl-questionnaire": "^4.2.2",
+    "@shikshalokam/sl-questionnaire": "2.3.0",
```
rajeevsathish, line 80: why are we downgrading?

src/app/client/src/app/modules/program-dashboard/pipes/chartType/chart-type.pipe.ts
```diff
@@ -11,6 +11,12 @@ export class ChartTypePipe implements PipeTransform {
         values: _.compact(chartData),
       });
     } else {
+      if(_.get(chartData, "filters")){
```
rajeevsathish, line 14: what will happen with this if the filters is not there?

5Amogh: Filters are mandatory for our use case here in the configuration still if they're not present this condition will not apply and other items will be returned as per the code

rajeevsathish (approved):

PR #8746, opened 2023-06-01
#ED-433 fix: Enabling graphs and tables for old programs

5Amogh:


src/app/client/src/app/modules/program-dashboard/components/program-datasets/program-datasets.component.ts
```diff
@@ -313,6 +313,7 @@ export class DatasetsComponent implements OnInit, OnDestroy {
     this.newData = false;
     this.noResult = false;
     this.districts = this.organisations = []
+    this.userAccess = true;
```
rajeevsathish, line 316: when is the value becomes false
this.userAccess

5Amogh: When the particular user is a program manager and that program has been enabled for program level reports (new feature)

5Amogh: after selecting a resource this is made true as both program manager and designer have access to resource/solution level reports and csvs

rajeevsathish (approved):

